PR #4, opened 2017-07-12
Enable use of Logger as a filter

jethrogb:
The spec parsing and matching of env_logger is very useful, but it will only let you log to stdout/stderr. With this change you can embed env_logger's logic in your own logger.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -162,6 +162,7 @@ mod filter;
 pub enum Target {
     Stdout,
     Stderr,
+    Silent,
 }
 
 /// The env logger.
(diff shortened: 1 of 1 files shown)


sebasmagri:
@jethrogb it looks good to me initially. However, could you please add an example or a few documentation lines explaining how to use this with your own logger?

jethrogb:
Added docs

sebasmagri:
Hi @jethrogb!

Could you please review and rebase changes against the latest master to merge this?

jethrogb:
Rebased

KodrAus:
Thanks for this @jethrogb! Sorry about the delay, I'll give this a proper review this week. You won't need to rebase again.

jethrogb:
> once the build is green

Fixed. Probably should've run `cargo build` after the rebase.

> without having to set a target explicitly

You don't *have* to set it. If you only call `EnvLogger::matches`, the target is never used.

KodrAus:
> You don't have to set it. If you only call EnvLogger::matches, the target is never used.

Ah you're right. Having a silent target is a prerequisite to anything else we'd want to do here anyway so there's no reason to block this on thinking about other ways to expose the filtering.


Thanks again!

PR #15, opened 2017-08-21
Catch up with changes to env_logger in rust-lang-nursery/log

mjkillough:
There were a few small changes to `env_logger` in rust-lang-nursery/log:

- rust-lang-nursery/log#196
- rust-lang-nursery/log#200

Without these changes, this crate does not compile.


sebasmagri:
@mjkillough thanks so much!

Could you please rebase against master again so we can merge it?

Regards,

mjkillough:
@sebasmagri - Done! I haven't rebased #16, but that should merge cleanly.

Issue #10, opened 2017-08-19
Add missing Errors and Panics section to documentation


sebasmagri:
- [ ] Missing error conditions on `Builder::try_init` and the root `try_init` function
- [ ] Missing _Panics_ section on `Builder::init` and the root `init` function

PR #16, opened 2017-08-21
A few improvements to the documentation

mjkillough:
This PR makes a bunch of changes to the documentation in order to fix:

- #11: Make reference to log crate into a hyperlink. Add an extra line to the summary at the beginning of the module-level documentation to indicate that this crate should be used with the standard logging facade (unlike something like `slog`).
- Consistently use monospaced font for `RUST_LOG`.
- #10: Add Errors/Panics section to `init()`/`try_init()` docs.
- #9: Avoid `unwrap()` in `Builder` doc string.
- #8: Provide examples for `Logger::new()` and `Logger:filter()`. Expand doc string for `Logger` to give more context and links to the typical ways of constructing it.

Please let me know if you'd like these changes as separate PRs - they are already separated by commit.

This PR needs the changes from #15 before it will compile or pass tests. (I've merged the two branches locally and confirmed both `cargo check`/`cargo doc` and `cargo test` produce the correct output).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -163,7 +163,26 @@ pub enum Target {
     Stderr,
 }
 
-/// The logger.
+/// The env logger.
+///
+/// This struct implements the `Log` trait from the [`log` crate][log-crate-url],
+/// which allows it to act as a logger.
+///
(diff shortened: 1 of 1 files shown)


sebasmagri:
Thanks!

Scutua (approved):

PR #18, opened 2017-08-22
Derive std traits for Target.

### LINKED ISSUE #6: Implement missing imps for Target
`Target` is missing the following traits:

- [ ] `Clone`
- [ ] `Copy`
- [ ] `Eq`
- [ ] `PartialEq`
- [ ] `Ord`
- [ ] `PartialOrd`
- [ ] `Hash`

mjkillough:
Fixes #6.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -157,7 +157,7 @@ mod filter;
 mod filter;
 
 /// Log target, either stdout or stderr.
-#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
+#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
 pub enum Target {
     Stdout,
     Stderr,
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -157,7 +157,7 @@ mod filter;
 mod filter;
 
 /// Log target, either stdout or stderr.
-#[derive(Debug)]
+#[derive(Clone, Copy, Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
```
KodrAus: Hmm, `Ord` and `PartialOrd` probably don't make sense for this `enum`. I think we should remove them, but the rest look good.

mjkillough: Sorry, I'd missed your comment on #6. I agree: I don't think they make sense, so I've removed them in the latest commit.

mjkillough:
@KodrAus - Thanks a lot for the review! Hopefully the latest commit addresses your comment. Please can you take another look?

sebasmagri:
Sure @KodrAus! Thanks!

Scutua (approved):

Issue #2, opened 2017-07-06
Output colors support?


sebasmagri:
As per [`log`'s review](https://internals.rust-lang.org/t/crate-evaluation-for-2017-05-16-log/5185). Check out `pretty_env_logger`.
[Issue in `log`](https://github.com/rust-lang-nursery/log/issues/37)

budziq:
I'd love such feature.

Unfortunately I see is no way to support color output in cross platform way without changing the architecture somewhat ([Builder::format](https://doc.rust-lang.org/log/env_logger/struct.Builder.html#method.format) having to return a String).
https://github.com/seanmonstar/pretty-env-logger shows colored output only on ANSI compliant terminals (as there you just need to add ansi control codes to the console). I've encountered this issue when working on https://github.com/azerupi/mdBook/pull/369

KodrAus:
We should be able to support basic colour support now that the output isn't a plain string.

I'll have a play with it using `term-painter` and see where it gets us. Are there any other recommendations for terminal colouring libs?

lnicola:
I've also seen https://github.com/BurntSushi/ripgrep/tree/master/termcolor and https://github.com/SergioBenitez/yansi.

KodrAus:
Hmm, `termcolor` looks interesting, thanks I'll dig into it! In classic @burntsushi style it looks like it solves all of life's problems.

KodrAus:
Hmm, so the whole `termcolor` API is pretty interesting. Looks like [`Buffer`s + `WriteColor`](https://docs.rs/termcolor/0.3.3/termcolor/struct.Buffer.html) could neatly solve this and #31 too, but means a new public dependency.

I'll spend some time spiking it out.

lnicola:
It seems pretty lightweight, with no extra dependencies except on Windows.

KodrAus:
Ah sorry I think half that comment stayed in my head.

If we want to support colours using `termcolor` then we'd need to change our format function from accepting a `Box<Write>` to accepting a `Box<WriteColor>` or something. In that case we're publicly exposing `termcolor`'s API and need to be careful about upgrading it. That might not actually be a problem though. Just something I want to keep in mind.

It does look like a pretty self-contained crate though, so if this works out I don't think I'd have a problem depending on it in `env_logger`.

BurntSushi:
[...]

PR #19, opened 2017-08-22
Avoid allocation of String during formatting.

### LINKED ISSUE #3: Avoid allocation of String in formatting
As per [`log`'s review](https://internals.rust-lang.org/t/crate-evaluation-for-2017-05-16-log/5185), formatting currently requires allocation via creating a `String`. This should be avoided.

mjkillough:
The format function is now passed a `&mut std::io::Write` trait object
and is responsible for outputting the formatted message to it. Format
functions can take care (by using the `std::fmt` macros) to avoid heap
allocations during formatting/outputting of log messages, and the
default format function takes advantage of this.

Alternatives considered:

 - Have the format function return a `fmt::Arguments` (from a call to
   `format_args!()`. Unfortunately I couldn't find a way to make this
   work with the lifetime parameter on `fmt::Arguments`.
 - Give the format function something like `fern`'s `FormatCallback`:
   https://docs.rs/fern/0.4.2/fern/struct.FormatCallback.html rather
   than giving it the `&mut Write`. This would be more future proof, but
   I worry it introduces unnecessary abstraction into `env_logger`,
   which generally tries to keep things simple.

... but I don't think either of those are better.

Some open questions:

 - Is referring to it as a 'format function' still appropriate? Should
   `Builder::format()` be renamed?
 - Should we be backwards-compatible with current env_logger by allowing
   people to give a `Fn(&Record) -> String` still? (Should we make
   upgrades easier or encourage people to rethink their format functions
   and whether they heap allocate?)

Fixes #3.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -186,7 +186,7 @@ pub enum Target {
 pub struct Logger {
     directives: Vec<Directive>,
     filter: Option<filter::Filter>,
-    format: Box<Fn(&Record) -> String + Sync + Send>,
+    format: Box<Fn(&mut Write, &Record) -> io::Result<()> + Sync + Send>,
     target: Target,
 }
 
@@ -203,12 +203,13 @@ pub struct Logger {
 /// extern crate env_logger;
 ///
 /// use std::env;
+/// use std::io;
 /// use log::{Record, LevelFilter};
 /// use env_logger::Builder;
 ///
 /// fn main() {
-///     let format = |record: &Record| {
-///         format!("{} - {}", record.level(), record.args())
+///     let format = |buf: &mut io::Write, record: &Record| {
+///         writeln!(buf, "{} - {}", record.level(), record.args())
 ///     };
 ///
 ///     let mut builder = Builder::new();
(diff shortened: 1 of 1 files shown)


mjkillough:
If it's desirable to add color support (#2) then perhaps we should pass some kind of new `Output` type to the format functions (rather than `std::io::Write`). This would allow us to add an `Output::supports_ansi_colors()` method. Formatters could use this in order to decide whether to use `ansi_term` to output colored logs on ANSI terminals, or to output non-colored logs for non-ANSI terminals.

We _could_ also build the machinery for outputting colors in a cross-platform way into the `Output` type, but that seems like massive scope creep. :)

sebasmagri:
@mjkillough color support is not in the current scope but something we could look at in the future. For now I think this is completely fine.

Thanks!

tailhook:
Hi,

Is there any actual need for this PR? The problem is that unbuffered writes of logs aren't very good if you have multiple threads or even multiple processes. I.e you will get a mess of different pieces of different log messages. Previous implementation added `\n` afterwards and even that was a problem. So I think the good idea is to make better buffering as added in https://github.com/rust-lang-nursery/log/pull/219 rather than remove buffering.

KodrAus:
I think this is a valid point. Thanks for following this up @tailhook! I'll open up an issue shortly to discuss sinks so we can iterate towards an implementation that satisfies all the things we expect from them.

PR #20, opened 2017-08-22
Debug traits for  structs

shivanth:
Fixes https://github.com/sebasmagri/env_logger/issues/14

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -145,6 +145,7 @@ use std::env;
 use std::io::prelude::*;
 use std::io;
 use std::mem;
+use std::fmt;
 
 use log::{Log, Level, LevelFilter, Record, SetLoggerError, Metadata};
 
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -459,6 +460,19 @@ impl Log for Logger {
     fn flush(&self) {}
 }
 
+impl fmt::Debug for Logger{
```
KodrAus: There are some nice [helper methods on `Formatter`](https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct) that can make this easier. So you could have something like:

```rust
f.debug_struct("Logger")
 .field("directives", &self.directives)
 .field("filter", &self.filter)
 .field("target", &self.target)
 .finish()
```

src/regex.rs
```diff
@@ -4,6 +4,7 @@ use std::fmt;
 
 use self::regex::Regex;
 
+#[derive(Debug)]
```
KodrAus: Do we want the debug representation to display the values of these private internal fields? Right now this will output the debug value of the `inner: Regex` field. We may well want that, but I think it's worth making that decision explicit. Are you ok with this @sebasmagri?

src/lib.rs
shivanth: Done :-)

src/regex.rs
shivanth: Any comments on this ?

KodrAus: I think this is ok. Let's merge this in.

PR #24, opened 2017-10-12
Update to the latest commit of `log` / deny debug impls, missing docs, and warnings.

jimmycuadra:
I was adding a deny attribute for missing debug impls, missing docs, and warnings, and discovered the crate doesn't compile against the latest API changes to `log`'s master branch, so I updated that too.


KodrAus:
Awesome, thanks @jimmycuadra!

Scutua (approved):

Scutua (approved):

Issue #23, opened 2017-09-24
Issues in log repo


KodrAus:
Tracking a few issues for `env_logger` that are still filed in `log`.

KodrAus:
https://github.com/rust-lang-nursery/log/issues/210

KodrAus:
https://github.com/rust-lang-nursery/log/issues/83

snowe2010:
https://github.com/rust-lang-nursery/log/issues/221

apiraino:
Hi @KodrAus are these issue being taken care of or tracked somewhere?

Specifically https://github.com/rust-lang/log/issues/210 is a longstanding known pitfall, still quite confusing before figuring that out.

thanks

PR #25, opened 2017-10-12
Adds a new constructor and init functions for custom env vars.

jimmycuadra:
Three new methods/functions that are like existing ones, but allow you to specify the name of the environment variable to parse:

* `Logger::from_env`, which is like `Logger::new`.
* `init_from_env`, which is like `init`.
* `try_init_from_env`, which is like `try_init`.

I also changed the docs for the init functions slightly to make it clearer what the difference is between them all when reading the docs for the crate root.

I haven't added any new tests, but I'd be happy to if desired. Let me know if there is a preferred way of working with environment variables to keep tests deterministic.

Refs #23
Refs rust-lang-nursery/log#83

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -448,9 +448,43 @@ impl Logger {
     /// [`init()`]: fn.init.html
     /// [`try_init()`]: fn.try_init.html
     pub fn new() -> Logger {
+        Self::from_env("RUST_LOG")
+    }
+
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks again @jimmycuadra!

I think this looks good. There probably isn't a whole lot of value testing the environment vars, since they just hand off to `parse` where the real work happens.

PR #27, opened 2017-10-16
Add examples for custom format and logger impls

KodrAus:
I think it would be good for use to build up a set of complete cargo examples. Here I've added two:

- Creating a custom logger (maybe we should make this one write to an in-memory `Vec<String>` or something?)
- Setting a custom format

These seem like good ones to start with because they use some new APIs we've introduced. We can use them to drive any changes we want to make to those APIs to make them a bit more ergonomic.

### DIFF
--- a/examples/custom_format.rs
+++ b/examples/custom_format.rs
@@ -1,0 +1,33 @@
+/*!
+Changing the default logging format.
+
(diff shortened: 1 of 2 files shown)


jimmycuadra:
Nice additions!

PR #32, opened 2017-10-31
Support colors and buffer writes

### LINKED ISSUE #31: Ensure log lines aren't interleaved
Make sure multiple calls to `Write::write` in log formats don't result in logs being interleaved when accessed concurrently.

KodrAus:
Closes #31 and #2 by using [`termcolor`](https://crates.io/crates/termcolor) to do the hard work.

# Colours

The default format now includes some colours:

colours

The colour API is still private because I'm not sure how we want to expose it.

# Buffering

This API goes back to buffering log records before printing them to the terminal, but we use a single buffer per thread rather than allocating for each log. I did some benchmarking and performance looks good overall.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,13 +148,15 @@ use std::env;
 use std::io::prelude::*;
 use std::io;
 use std::mem;
-use std::fmt;
 use std::cell::RefCell;
 
 use log::{Log, LevelFilter, Level, Record, SetLoggerError, Metadata};
-use termcolor::{ColorChoice, Color, ColorSpec, BufferWriter, Buffer, WriteColor};
+use termcolor::{ColorChoice, Color, BufferWriter};
 
 pub mod filter;
+pub mod fmt;
+
+use self::fmt::Formatter;
 
 /// Log target, either stdout or stderr.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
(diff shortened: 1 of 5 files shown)


src/lib.rs
```diff
@@ -183,9 +186,24 @@ pub enum Target {
...
+    format: Box<Fn(&mut Formatter, &Record) -> io::Result<()> + Sync + Send>,
+}
+
+/// A formatter to write logs into.
+/// 
+/// `Formatter` implements the standard [`Write`] trait for writing log records.
+/// It also supports terminal colors, but this is currently private.
+/// 
+/// [`Write`]: https://doc.rust-lang.org/stable/std/io/trait.Write.html
+pub struct Formatter {
```
KodrAus: I think I'll move this into a new `fmt` module, since when we do make colouring public there will probably be some more types to go along with it.

If you just create your closure as part of the `.format` call then you won't need to name the `Formatter` type directly.

src/lib.rs
```diff
@@ -442,21 +473,88 @@ impl Log for Logger {
...
+                let mut tl_buf = tl_buf.borrow_mut();
+
+                if tl_buf.is_none() {
+                    *tl_buf = Some(Formatter::new(self.writer.buffer()));
+                }
+
+                // The format is guaranteed to be `Some` by this point
+                let mut formatter = tl_buf.as_mut().unwrap();
+
+                let _ = (self.format)(&mut formatter, record)
```
KodrAus: I think this should clarify its error conditions a bit better:

- We only print the record to the terminal if the record is successfully formatted (which it pretty much always should)
- We always clear the buffer even if the record wasn't printed. Seems like it might not be ideal, but I'm not really sure what would cause a print to fail.

I think it needs a little more thought and should be refactored to clarify the intent a bit better.

src/lib.rs
```diff
@@ -442,21 +473,88 @@ impl Log for Logger {
...
+    }
+}
+
+impl<W> Write for StyledFormatter<W>
+    where W: WriteColor
+{
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        self.buf.set_color(&self.spec)?;
+        let write = self.buf.write(buf);
+        self.buf.reset()?;
```
KodrAus: Not returning the bytes written if the terminal fails to reset seems like a bad thing to do.

KodrAus:
Ok I'm happy. Docs can be dealt with down the track.

@BurntSushi if you've got a few minutes sometime I'd be grateful if you could give this a quick scan and let me know if anything looks totally insane :)

BurntSushi:
@KodrAus LGTM! Nice work. :-)

PR #33, opened 2017-11-01
Add rfc3339 timestamps to default format

### LINKED ISSUE #30: Include timestamp on default format
We should include an ISO-formatted timestamp on the default log format. It's not uncommon for apps running in virtualised or headless environments to capture stdout/stderr for later review, and having timestamps is a helpful bit of insight.

Since formats are configurable, tools like cargo that are probably being watched can opt-out of displaying a timestamp to reduce noise.

KodrAus:
Closes #30 

Adds [RFC3339](https://tools.ietf.org/html/rfc3339) formatted timestamps to the default log format:

timestamps

I've made an effort to avoid allocating temporary strings or needing to parse a format for each log, so there's a currently private `Timestamp` struct that can be pulled off a `Formatter` that does this.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ publish = false # this branch contains breaking changes
 log = { git = "https://github.com/rust-lang-nursery/log.git", features = ["std"] }
 regex = { version = "0.2", optional = true }
 termcolor = "0.3"
+chrono = "0.4"
 
 [[test]]
 name = "regexp_filter"
--- a/README.md
+++ b/README.md
@@ -95,11 +95,11 @@ $ RUST_LOG=my_lib=info cargo test
      Running target/debug/my_lib-...
 
 running 2 tests
-INFO:my_lib::tests: logging from another test
-INFO:my_lib: add_one called with -8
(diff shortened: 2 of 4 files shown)


mjkillough:
@KodrAus - Apologies for the late review, I was on holiday until this afternoon.

This looks great! :) I didn't realise how great the formatting in `chrono` was.

tailhook:
Well, since the UTC time is used anyway (which is a good thing), I think `+00:00` is excessive, `Z` is usually used in that case and is clear enough and shorter (so easier to observe in console, since this is the primary target of env_logger). Also, I think that nanosecond precision, while can be occasionally useful is also excessive. I'm usually okay with second granularity. May be millisecond if possible.

In my opinion `time::now_utc().rfc3339()` is good enough and produces value that looks like: `2017-01-01T00:00:00Z`, and also avoids few more dependencies.

KodrAus:
Thanks @mjkillough and @tailhook!

The `time` crate is effectively deprecated in favour of `chrono` so I don't think we should depend on it directly. I think the points about the format are good, we might end up having to reconstruct a simpler rfc3339 format manually though. But that's not the end of the world.

KodrAus:
Added #34 to sort the format out.

Issue #34, opened 2017-11-06
Reduce noise in the timestamp format


KodrAus:
Was raised in #33 that the current timestamp format is pretty verbose, but could be simpler while still being useful and valid RFC3339. We should:

- Remove nanoseconds
- Replace `+00:00` with a simple `Z`

KodrAus:
The current implementation uses `chrono`'s formatting `Item`s to avoid allocating temporary strings or needing to parse the same format each time a record is logged. It's simple because there's only 1 `Item` to deal with. If we need to reconstruct an RFC3339 format then it'll get more complex to build the format this way.

tailhook:
> The time crate is effectively deprecated in favour of chrono so I don't think we should depend on it directly

Well, I think this is where practicality beats purity. Since `env_logger` doesn't expose chrono's structures anywhere, it's okay to depend on `time` because it's easy to opt-out at any time.

mjkillough:
In #35 I've attempted to make the timestamp format configurable. I wasn't able to alias `+00:00` to `Z` with this approach, but I think that should be possible with some extra changes to `chrono`. (I think it's probably worth making those changes to `chrono` anyway).

PR #35, opened 2017-11-06
Allow custom timestamp formats.

mjkillough:
Past experience has taught me that you'll never have one logging time
format that pleases everyone, so perhaps we should just allow it to be
configurable. :)

We can use `StrftimeItems` to parse the format string once when we build
the `Logger`. We limit the lifetime of the format string to `'static` as
it keeps things simple. (If we wanted to accept arbitrary lifetimes,
I think we'd either need to introduce a lifetime parameter on `Logger`,
or have a self-referential struct somewhere).

This commit also removes the nanoseconds, as suggested in #34. It does
not alias `+00:00` to `+Z`, as there doesn't seem to be a way to do this
with `chrono::format::strftime`. (It is possible to do it by
constructing our own iterator of formatting `Item`s, but I think
`chrono` should learn to allow this via `StftimeItems` instead).

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,11 +95,11 @@ $ RUST_LOG=my_lib=info cargo test
      Running target/debug/my_lib-...
 
 running 2 tests
-INFO: 2017-11-01T21:37:57.073523133+00:00: my_lib::tests: logging from another test
-INFO: 2017-11-01T21:37:57.073523133+00:00: my_lib: add_one called with -8
+INFO: 2017-11-01T21:37:57+00:00: my_lib::tests: logging from another test
+INFO: 2017-11-01T21:37:57+00:00: my_lib: add_one called with -8
 test tests::it_handles_negative_numbers ... ok
-INFO: 2017-11-01T21:37:57.073523133+00:00: my_lib::tests: can log from the test too
-INFO: 2017-11-01T21:37:57.073523133+00:00: my_lib: add_one called with 2
+INFO: 2017-11-01T21:37:57+00:00: my_lib::tests: can log from the test too
+INFO: 2017-11-01T21:37:57+00:00: my_lib: add_one called with 2
 test tests::it_adds_one ... ok
 
 test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured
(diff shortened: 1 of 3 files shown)


mjkillough:
Er, whoops. This doesn't actually do what I intended it to do, as it does parse the format string each time a log message is formatted.

I think I meant to drain the `StrftimeItems` iterator into a `Vec` when building the logger, then pass a copy of that `Vec` to each `Formatter`. It would mean an extra allocation when each thread first logs, but shouldn't involve any allocations for subsequent log messages.

KodrAus:
Thanks @mjkillough!

I think we'll want to think about the question of supporting multiple formats eventually and it might look a lot like this, but I'm a bit nervous about leaking `chrono` as a dependency in this way on the outset.

We can get the default more compact RFC3339 format by manually constructing a slice of items, something like:

```rust
use chrono::format::Item;

const ITEMS: &'static [Item<'static>] = {
    use chrono::format::Item::*;
    use chrono::format::Numeric::*;
    use chrono::format::Fixed::*;
    use chrono::format::Pad::*;

    &[
        Numeric(Year, Zero),
        Literal("-"),
        Numeric(Month, Zero),
        Literal("-"),
        Numeric(Day, Zero),
        Literal("T"),
        Numeric(Hour, Zero),
        Literal(":"),
        Numeric(Minute, Zero),
        Literal(":"),
        Numeric(Second, Zero),
        Fixed(TimezoneOffsetZ),
    ]
}
```

It's not exactly pretty, but it's simple and means we can keep all the formatting localised to the `fmt` module for now, and think about an API for configuring the default format without having to redefine it later. Things like colours might also factor into that.

What do you think?

Thanks again for tackling this!

KodrAus:
It turns out the gymnastics I went through in the last PR for making a custom iterator were also not necessary. We can just use `ITEMS.iter().cloned()` when `ITEMS: &'static [Item<'static>]` :)

mjkillough:
Thanks a lot for the quick response!

> I'm a bit nervous about leaking `chrono` as a dependency in this way on the outset.

Yeah, that's fair. I sat down to do this because I was curious about how difficult it would be, but I agree it might be worth deferring this type of customization until later.

I had hoped we could expose it as "`strftime`/`strptime` style formatting" and hide the fact that `chrono` is doing it. However, it seems there are a few quirks in how `chrono` implements this, which may mean we're really leaking it as a dependency anyway.

> Things like colours might also factor into that.

That's a great point - I hadn't considered that.

> We can get the default more compact RFC3339 format by manually constructing a slice of items, something like:

Yeah, I think that is best for now. It's certainly concise as a slice, and it does allow us to use `TimezoneOffsetZ` (which didn't seem possible with `StrftimeItems`).

I'll close this. :)

PR #36, opened 2017-11-09
Make the timestamp format more compact

### LINKED ISSUE #34: Reduce noise in the timestamp format

KodrAus:
Fixes #34 

format

Makes the RFC3339 timestamp format more compact. This is still just using a hardcoded set of `chrono::format::Item`s. I think we'll definitely want to make this configurable in the future somehow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,7 @@ environment variable that corresponds with the log messages you want to show.
 
 ```bash
 $ RUST_LOG=info ./main
-starting up
+INFO: 2017-11-09T02:12:24Z: main: starting up
 ```
 
 ### In tests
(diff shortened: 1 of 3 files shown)


KodrAus:
Oops! I forgot to update our example formatted log records.

tailhook:
Looks great. I don't think a configurable format is much needed. 

Any blockers for a release?

KodrAus:
@tailhook We're just waiting on an `0.4` release of `log` and then this should be ready to go. I'll create an issue for running through the docs and give them a polish but that doesn't need to block a release.

PR #39, opened 2017-12-07
Upgrade to log 0.4.0-rc.1

KodrAus:
- Fixes up the build after some changes to `log`
- Targets the new RC version of `log`
- Specifies an RC version for `env_logger`

The goal would be to get this up on crates.io sooner rather than later so we can loosely follow `log`'s own path to a stable `0.4` release.


KodrAus:
That's what I get for not running tests before committing...

PR #41, opened 2017-12-07
Pad log level so output is more aligned

killercup:
This makes the example print this:

```
TRACE: 2017-12-07T13:44:04Z: default: some trace log
DEBUG: 2017-12-07T13:44:04Z: default: some debug log
 INFO: 2017-12-07T13:44:04Z: default: some information log
 WARN: 2017-12-07T13:44:04Z: default: some warning log
ERROR: 2017-12-07T13:44:04Z: default: some error log
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -249,7 +249,7 @@ impl Builder {
                     Level::Error => Color::Red,
                 };
 
-                let write_level = write!(buf.color(level_color), "{}:", level);
+                let write_level = write!(buf.color(level_color), "{:>5}:", level);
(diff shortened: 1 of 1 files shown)


euclio:
I personally prefer the padding on the right side. That's what [pretty-env-logger](https://github.com/seanmonstar/pretty-env-logger) does.

KodrAus:
Thanks @killercup! I think aligning the level to the colon is totally fine. I personally agree with @euclio but can't think of a reason why it would be universally better to align the other way.

Issue #42, opened 2017-12-10
Make the color API public


KodrAus:
There's a private API for writing colours that we use in the default format. This should be public so anyone building their own format can use it. Colours currently use an enum from `termcolor`, which means we'd need to make it a public dependency, or wrap it somehow.

PR #46, opened 2017-12-22
Public color API

### LINKED ISSUE #45: Could color be disabled?
I really like the new format, including the color!

Unfortunately when the log output is piped and stored elsewhere (for example in AWS CloudWatch) I see all the control characters which makes reading the log actually harder.

Could the color be disabled with a feature flag?

KodrAus:
Closes #45 and #42

# Some thoughts

- Do we want to treat new environment variables as breaking changes? Users overriding environment variables would probably want to know when new ones are added so they can be overridden too
- Should `Env` be an `&mut self` builder like the others?
- Should `Style` implement `Write` so you can call `writeln!(read_style, "{}", record.args())` to write an entire line using styles?

# Refactorings

I've moved all the terminal writing related code, like `Target` and the new `WriteStyle` into the `fmt` module to try and encapsulate it a bit better. Now we don't depend on `termcolor` directly in the `Logger` or `Builder`.

# Disabling colors

Adds a new builder property for whether or not to include colors. By default this is `RUST_LOG_STYLE`, but it can be overridden.

Setting this environment variable to `never` will disable colors and other styles:

```shell
$ export RUST_LOG_STYLE=never
$ ./my-app
```

Valid values are:

- `auto` (or missing/invalid) will decide whether or not the terminal supports colors
- `always` will always use colors
- `never` will never use colors

In order to support multiple environment variables, I've refactored our `from_env` functions to accept a generic `T: Into<Env>`, where `Env` is a container for the environment variables we care about. These methods can now be called in a few ways:

```rust
// reads filters from `MY_LOG` and styles from `RUST_LOG_STYLE`
env_logger::init_from_env("MY_LOG");

// reads filters from `MY_LOG` and styles from `MY_LOG_STYLE`
env_logger::init_from_env(Env::default().filter("MY_LOG").write_style("MY_LOG_STYLE"));
```

This lets us add new environment variables in the future without potentially breaking people. But it does mean if you're overriding all environment variables that new ones could slip in without you noticing.

# Using alternative environment variables

Since we use two environment variables to configure the logger we need an ergonomic way to pass different combinations of those variables to `from_env` methods. This PR adds an `Env` type with builder methods for naming environment variables:

```rust
env_logger::init_from_env(Env::new().filter("MY_LOG"));
```

With a few `From` conversions, the above is also equivalent to:

```rust
env_logger::init_from_env("MY_LOG");
```

Whether or not we want to keep these conversions is up for discussion.

# Writing colors

The color API has been refactored so it's a bit nicer to use in your own formats:

```rust
let mut style = buf.style();

style.set_color(Color::Red).set_bold(true).set_bg(Color::White);

writeln!(buf, "{}", style.value(42))
```

This saves you from having to split the writes into multiple calls and juggle `Result` types.

# Writing timestamps

Oh the `Formatter.timestamp` method is now also public so you can grab timestamps.

### DIFF
--- a/examples/custom_format.rs
+++ b/examples/custom_format.rs
@@ -7,10 +7,11 @@ Before running this example, try setting the `MY_LOG_LEVEL` environment variable
 $ export MY_LOG_LEVEL='info'
 ```
 
-Also try setting the `MY_LOG_STYLE` environment variable to `0` to disable colors:
+Also try setting the `MY_LOG_STYLE` environment variable to `never` to disable colors
+or `auto` to enable them:
 
 ```no_run,shell
-$ export MY_LOG_STYLE=0
+$ export MY_LOG_STYLE=never
 ```
 
 If you want to control the logging output completely, see the `custom_logger` example.
--- a/examples/custom_format.rs
+++ b/examples/custom_format.rs
@@ -17,12 +17,19 @@ extern crate env_logger;
 use std::env;
 use std::io::Write;
 
+use env_logger::{Builder, fmt};
+
 fn init_logger() {
-    let mut builder = env_logger::Builder::new();
+    let mut builder = Builder::new();
 
     // Use a different format for writing log records
     builder.format(|buf, record| {
-        writeln!(buf, "My formatted log: {}", record.args())
+        let mut style = buf.style();
+        style.set_bg(fmt::Color::Yellow).set_bold(true);
+
+        let timestamp = buf.timestamp();
+
+        writeln!(buf, "My formatted log ({}): {}", timestamp, style.value(record.args()))
     });
 
     if let Ok(s) = env::var("MY_LOG_LEVEL") {
--- a/examples/custom_format.rs
+++ b/examples/custom_format.rs
@@ -4,13 +4,13 @@ Changing the default logging format.
 Before running this example, try setting the `MY_LOG_LEVEL` environment variable to `info`:
 
 ```no_run,shell
-$ export MY_LOG_LEVEL = 'info'
+$ export MY_LOG_LEVEL='info'
 ```
 
 Also try setting the `MY_LOG_STYLE` environment variable to `0` to disable colors:
 
 ```no_run,shell
-$ export MY_LOG_STYLE = 0
+$ export MY_LOG_STYLE=0
 ```
 
 If you want to control the logging output completely, see the `custom_logger` example.
--- a/examples/custom_format.rs
+++ b/examples/custom_format.rs
@@ -20,7 +20,6 @@ If you want to control the logging output completely, see the `custom_logger` ex
 extern crate log;
 extern crate env_logger;
 
-use std::env;
 use std::io::Write;
 
 use env_logger::{Env, Builder, fmt};
--- a/examples/custom_format.rs
+++ b/examples/custom_format.rs
@@ -7,6 +7,12 @@ Before running this example, try setting the `MY_LOG_LEVEL` environment variable
 $ export MY_LOG_LEVEL = 'info'
 ```
 
+Also try setting the `MY_LOG_STYLE` environment variable to `0` to disable colors:
+
+```no_run,shell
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -238,31 +256,50 @@ impl Builder {
...
 
+    /// Initializes the log builder from the environment.
+    pub fn from_env<'a, E>(env: E) -> Self
+    where
+        E: Into<Env<'a>>
+    {
+        let mut builder = Builder::new();
+        let Env { filter, write_style } = env.into();
+
+        if let Ok(s) = env::var(&*filter) {
```
KodrAus: Repeating this check is a bit nasty. I think we should add methods to `Env` to get an `Option` for each variable

src/lib.rs
```diff
@@ -482,13 +555,13 @@ impl Log for Logger {
                 let mut tl_buf = tl_buf.borrow_mut();
 
                 if tl_buf.is_none() {
-                    *tl_buf = Some(Formatter::new(self.writer.buffer()));
+                    *tl_buf = Some(Formatter::new(self.writer.buffer(), self.write_style));
```
KodrAus: This is problematic because multiple loggers working on the same thread with different configurations will get unexpected results. The problem is the same with the `Buffer`, since it's tied to the `Logger`.

I'm thinking we just prevent you building a `Logger` directly and require it's configured globally, but am open to other suggestions.

KodrAus:
Looks like we could handle the colors better by using `termcolor`'s `ColorChoice` enum better. That way we won't need to thread the extra state around.

KodrAus:
We should align the values of `RUST_LOG_STYLE` with Cargo's `--color` inputs.

KodrAus:
I've done some refactoring so all the code around writing to the terminal is contained in the `fmt` module. This means the root `Logger` and `Builder` are mostly delegating to the `filter` and `fmt` modules to work.

I think this will make the library easier to navigate and use in the future since things are getting more complex in `env_logger`.

examples/custom_format.rs
```diff
@@ -4,7 +4,13 @@ Changing the default logging format.
...
 
 ```no_run,shell
-$ export MY_LOG_LEVEL = 'info'
+$ export MY_LOG_LEVEL='info'
+```
+
+Also try setting the `MY_LOG_STYLE` environment variable to `0` to disable colors:
+
+```no_run,shell
+$ export MY_LOG_STYLE=0
```
KodrAus: This value is wrong. It should be `auto`, `always` or `never`.

examples/custom_logger.rs
```diff
@@ -4,7 +4,13 @@ Using `env_logger` to drive a custom logger.
...
 
 ```no_run,shell
-$ export MY_LOG_LEVEL = 'info'
+$ export MY_LOG_LEVEL='info'
+```
+
+Also try setting the `RUST_LOG_STYLE` environment variable to `0` to disable colors:
+
+```no_run,shell
+$ export RUST_LOG_STYLE=0
```
KodrAus: Same as other style examples.

examples/default.rs
```diff
@@ -4,7 +4,13 @@ Using `env_logger`.
...
 
 ```no_run,shell
-$ export MY_LOG_LEVEL = 'info'
+$ export MY_LOG_LEVEL='info'
+```
+
+Also try setting the `RUST_LOG_STYLE` environment variable to `0` to disable colors:
+
+```no_run,shell
+$ export RUST_LOG_STYLE=0
```
KodrAus: Same as other style examples.

src/fmt.rs
```diff
@@ -9,99 +9,473 @@
...
+    Never,
+}
+
+impl Default for WriteStyle {
+    fn default() -> Self {
+        WriteStyle::Auto
+    }
+}
+
+pub(crate) struct Writer(BufferWriter);
```
KodrAus: Even though these are private right now we should add some docs to them.

Issue #43, opened 2017-12-10
Release 0.5.0


KodrAus:
Now that `log` has an `0.4.0` release out we should get `env_logger` `0.5.0` out. Before that, we should sort out:

- [ ] #13
- [ ] #42

KodrAus:
I've published a [new pre-release](https://docs.rs/crate/env_logger/0.5.0-rc.2) with the upgrade to `log 0.4.0`, some tweaks to the default format and the public colour API. I'd be pretty happy to call this `0.5.0` as-is and work through the rest of those issues as non-breaking changes, if people want to give it a go and let us know what you think!

ignatenkobrain:
I would love to see 0.5.0 release because tokei now depends on rc1 and we try to avoid packaging unreleased versions of crate. Just my 5Kč.

KodrAus:
Yeh, I don't want to be sitting on the pre-release too long so will try to get a stable release out ASAP. I'm being a bit cautious though becuase `env_logger` is pretty widely used.

kpcyrd:
@KodrAus I'd love to have #47 in 5.0.0, but I think it's still possible to add this in a followup release if needed.

KodrAus:
@kpcyrd I think we can solve #47 and the other issues mentioned there in patches after we release `0.5`.

I'm just going to go over the docs today and release `0.5`, so people upgrading to `log` `0.4` get the best experience.

KodrAus:
I've released `0.5.0` [on crates.io](https://crates.io/crates/env_logger) now.

We can continue working on these other features in non-breaking changes. I'll put together some release notes, which will include all the notes from the pre-releases.

Thanks everyone who helped make this release possible!

PR #49, opened 2018-01-03
Dependencies: log: 0.4.0-rc.1 -> 0.4.0

matthiasbeyer:
Tests run fine for me... so I assume we can update `log` to the `0.4.0` release, can't we?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ categories = ["development-tools::debugging"]
 keywords = ["logging", "log", "logger"]
 
 [dependencies]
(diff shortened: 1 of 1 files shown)


matthiasbeyer:
Btw: Part of #43

KodrAus:
Thanks @matthiasbeyer! Looks like everything still works nicely with the proper `0.4.0` release.

I'd like to get #46 merged before we do another pre-release of `env_logger` (there's a lot of new API surface area), but in the meantime you should be able to use a `git` dependency if you've run into any cargo issues with `log 0.4.0-rc1` vs `log 0.4.0`.

PR #52, opened 2018-01-10
Update version in README

matthiasbeyer:
I also played with [version-sync](https://github.com/mgeisler/version-sync) to automatically detect these errors, but it does not work with the `[dev-dependencies]` section in the readme... so no luck there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,10 +51,10 @@ Tests can use the `env_logger` crate to see log messages generated during that t
 
 ```toml
 [dependencies]
-log = "0.4.0-rc.1"
+log = "0.4.0"
 
 [dev-dependencies]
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks @matthiasbeyer! Hmm, maybe `version-sync` is only looking for [either a `dependencies` or a `dev-dependencies` block in a single snippet](https://github.com/mgeisler/version-sync/blob/master/src/lib.rs#L165-L166), so one of them is being ignored in our README?

I'll merge this in, but if you'd like to play with it some more I'd be happy if the whole README got reformatted to improve our automation.

matthiasbeyer:
From what I currently see is that `version-sync` simply does not support `dev-dependencies`... so there's nothing I can do there.

PR #53, opened 2018-01-14
Allow building a Logger yourself

KodrAus:
Previously, I removed the ability to build and get a hold of a `Logger` type, because it relies on thread-local state that was configured using the style options of the first `Logger` to write a record.

This PR fixes up that situation so multiple `Logger`s can log on the same thread using the styles they were originally configured with. Each time we access the thread-local buffer we check to see if its configuration matches the current `Logger`. If it doesn't then we drop it and create a new one.

### DIFF
--- a/examples/direct_logger.rs
+++ b/examples/direct_logger.rs
@@ -1,0 +1,40 @@
+/*!
+Using `env_logger::Logger` and the `log::Log` trait directly.
+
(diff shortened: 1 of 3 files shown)


KodrAus:
I've also added an example of using a `Logger` to write records directly, without using the `log` macros. It also works as a smoke test for the changes in this PR.

PR #57, opened 2018-01-17
Avoid panicking for logging-while-logging

alexcrichton:
The `borrow_mut` panic could trip if a log happened while otherwise already
logging, so this adds a test and then avoid the panic in a relatively naive way.
If the panic should be avoided differently though, just let me know!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,5 +23,9 @@ chrono = "0.4"
 name = "regexp_filter"
 harness = false
 
+[[test]]
+name = "log-in-log"
+harness = false
+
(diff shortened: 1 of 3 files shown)


alexcrichton:
Thanks! If possible, mind releasing an 0.5.1 as well? If you'd prefer I can also run the test suite of rust-lang/rust against 0.5.0 with this patch to make sure it works first too :)

KodrAus:
No problem! I've pushed out `0.5.1`, it's a useful patch in its own right too :+1:

PR #60, opened 2018-01-18
Use atty to check for color support

### LINKED ISSUE #56: Detect type of output to omit coloring in redirected output.
May be it is not so clear title, but

On Windows with default formatter, colors are enabled. When running some app with `env_logger` inside console, it outputs colored messaged. 
So it does (tries to do) when you redirect your app to a file (via `app.exe 2> app.log`) and that file becomes pretty messy (full of lines like `[m[32m INFO[m 2018-01-17T14:56:26Z`).

I think it would be worth to detect whether STDERR is a tty (console in terms of Windows) or a file.

I am not sure whether it is an issue of `env_logger` or `termcolor`, but `ripgrep`, for instance, behaves correctly. So may be it can be done here.

quadrupleslap:
Is this all that's needed to close #56 ?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ log = { version = "0.4.0", features = ["std"] }
 regex = { version = "0.2", optional = true }
 termcolor = "0.3"
 chrono = "0.4"
+atty = "0.2"
 
 [[test]]
 name = "regexp_filter"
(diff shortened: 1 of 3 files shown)


KodrAus:
Thanks @quadrupleslap! This looks good to me. I'm just on mobile at the moment but we should be able to test this by piping the default example to files and things, and making sure the colour control characters aren't written.

KodrAus:
Just gave this a quick check and it looks like it does what we need!

Thanks again @quadrupleslap!

PR #66, opened 2018-02-13
Allow toggling parts of the default format

### LINKED ISSUE #48: Make it easy to define a slightly different format (without writing a formatter from scratch)
Hi there! 

While it is possible to define a custom formatting via the `Builder`, often one only wants to adjust small things. For example, the timestamp is way too noisy for my application, so I want to disable it. But now I have to do everything else by myself. Especially right now I cannot use colors, so I cannot implement the formatting I want. It would be nice to say something like `builder.with_timestamps(false)`. Sure, this would introduce one (or a few, if we want to add options for other parts -- like the module path -- as well) boolean flag that has to be checked for each log message. But that should be fine, right?

KodrAus:
closes #48 

# `default_format` methods

This PR tweaks the default format a bit so we can conditionally enable/disable parts of it in the main `Builder`. The API looks like:

```rust
impl Builder {
    fn default_format(&mut self) -> &mut Self;
    fn default_format_timestamp(&mut self, write: bool) -> &mut Self;
    fn default_format_module_path(&mut self, write: bool) -> &mut Self;
    fn default_format_level(&mut self, write: bool) -> &mut Self;
}
```

This isn't the most elegant API, but I think it solves the problem of conveniently tweaking the default format without needing to import types or figure out the custom formatting API. We could bikeshed the naming (maybe `use_default_format` and `write_default_format_timestamp`?)

Calling `default_format_*` will store whether or not we want to include those parts of the log, without affecting a custom format. That means:

```rust
Builder::new()
    .format(|buf, record| { ... })
    .default_format_timestamp(false)
    .init();
```

Is the same as:

```rust
Builder::new()
    .format(|buf, record| { ... })
    .init();
```

Setting a custom format won't clobber any values passed to `default_format_*` methods. That means:

```rust
Builder::new()
    .default_format_timestamp(false)
    .format(|buf, record| { ... })
    .default_format()
    .init();
```

Is the same as:

```rust
Builder::new()
    .default_format_timestamp(false)
    .init();
```

I'm not expecting users to hop between custom and default formats, but I thought these were the least surprising semantics.

The docs need fleshing out.

# `from_default_env`

This PR also adds a new `from_default_env` method that's a convenient way of calling `from_env(env_logger::Env::default())`.

### DIFF
--- a/examples/custom_default_format.rs
+++ b/examples/custom_default_format.rs
@@ -25,7 +25,8 @@ use env_logger::{Env, Builder};
 
 fn init_logger() {
     let env = Env::default()
-        .filter("MY_LOG_LEVEL");
+        .filter("MY_LOG_LEVEL")
+        .write_style("MY_LOG_STYLE");
 
     let mut builder = Builder::from_env(env);
 
--- a/examples/custom_default_format.rs
+++ b/examples/custom_default_format.rs
@@ -1,0 +1,43 @@
+/*!
+Disabling parts of the default format.
+
+Before running this example, try setting the `MY_LOG_LEVEL` environment variable to `info`:
+
+```no_run,shell
+$ export MY_LOG_LEVEL='info'
+```
+
+Also try setting the `MY_LOG_STYLE` environment variable to `never` to disable colors
+or `auto` to enable them:
+
+```no_run,shell
+$ export MY_LOG_STYLE=never
+```
+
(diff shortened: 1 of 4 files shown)


KodrAus:
Sorry @LukasKalbertodt I meant to cc you, but I started writing this PR on one day and actually opened it on the next and forgot to!

PR #67, opened 2018-02-17
Use humantime instead of chrono

tailhook:
I've just added rfc3339 timestamp support in humantime. And I find it's good match for env_logger.

Performance difference (just time formatting not logger):
```
     Running target/release/deps/datetime_format-7e6d1b8a3104a052

running 2 tests
test rfc3339_chrono            ... bench:         730 ns/iter (+/- 21)
test rfc3339_humantime_seconds ... bench:          73 ns/iter (+/- 3)
```
This uses same definition of chrono formatter as in env_logger.

Binary size difference:
```
- 0.2%   1.6%  10.5KiB chrono
- 0.0%   0.0%     305B time
+ 0.0%   0.3%   1.7KiB humantime
```
Measured using cargo-bloat on `examples/default.rs`

Dependencies difference:
```
 ├── atty v0.2.6
 │   [dependencies]
 │   └── libc v0.2.36
-├── chrono v0.4.0
+├── humantime v1.1.0-beta.3
 │   [dependencies]
-│   ├── num v0.1.42
-│   │   [dependencies]
-│   │   ├── num-integer v0.1.36
-│   │   │   [dependencies]
-│   │   │   └── num-traits v0.2.0
-│   │   ├── num-iter v0.1.35
-│   │   │   [dependencies]
-│   │   │   ├── num-integer v0.1.36 (*)
-│   │   │   └── num-traits v0.2.0 (*)
-│   │   └── num-traits v0.2.0 (*)
-│   └── time v0.1.39
-│       [dependencies]
-│       └── libc v0.2.36 (*)
-│       [dev-dependencies]
-│       └── winapi v0.3.4
+│   └── quick-error v1.2.1
 ├── log v0.4.1
 │   [dependencies]
 │   └── cfg-if v0.1.2

```
Reported by `cargo tree`

None of that is actually a life-saver. But there is [at least one crate](https://github.com/BurntSushi/ripgrep/pull/772/commits/e633d5062f662268fad1dd5d3c86e2060a6984e1) that avoids using env_logger specifically because it depends on quite large dependency tree of chrono.

So idea is to make `humantime` the time parser/formatter specifically for situations where chrono is an overkill. Humantime only depends on quick-error for error boilerplate and works with stdlib's SystemTime and Duration.

What do you think?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,7 @@ keywords = ["logging", "log", "logger"]
 log = { version = "0.4.0", features = ["std"] }
 regex = { version = "0.2", optional = true }
 termcolor = "0.3"
-chrono = "0.4"
+humantime = "1.1.0"
 atty = "0.2"
 
 [[test]]
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -1,34 +1,34 @@
 //! Formatting for log records.
-//! 
+//!
 //! This module contains a [`Formatter`] that can be used to format log records
 //! into without needing temporary allocations. Usually you won't need to worry
 //! about the contents of this module and can use the `Formatter` like an ordinary
 //! [`Write`].
-//! 
+//!
 //! # Formatting log records
-//! 
+//!
 //! The format used to print log records can be customised using the [`Builder::format`]
 //! method.
 //! Custom formats can apply different color and weight to printed values using
 //! [`Style`] builders.
-//! 
+//!
 //! ```
 //! use std::io::Write;
 //! use env_logger::fmt::Color;
-//! 
+//!
 //! let mut builder = env_logger::Builder::new();
-//! 
+//!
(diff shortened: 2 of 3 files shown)


KodrAus:
Thanks @tailhook! This looks like a total improvement to me so I'm inclined to merge this in once I get a chance to sit down and dig into it later.

tailhook:
Done. I've force-pushed the branch, but the only change here is a new `humantime` version of 1.1.0.

KodrAus:
Thanks!

@alexcrichton would there be any trouble for `rustc` if we change `env_logger`'s dependencies in a patch version? It has some kind of whitelist it uses, right? This PR replaces `chrono` with `humantime`, which itself only depends on `quick-error`.

alexcrichton:
Nah it should be totally fine for rustc!

Issue #72, opened 2018-02-27
Allow disabling the humantime dependency


briansmith:
Some environments, e.g. Kubernetes, already timestamp every log message. Thus, adding our own timestamp if wastefully duplicative. We currently have a custom formatter to disable the timestamping.

That works OK, as far as the runtime characteristic of the application are concerned. However, because Cargo downloads each dependency serially, and because Cargo builds each dependency serially, every single dependency inhibits the build performance by some constant factor that's independent of the complexity of the dependency. Some applications have continuous integration and/or container build systems that have to rebuild everything from scratch any time any dependency changes, and/or on every commit, so minimizing the from-scratch download + build time is important.

I noticed when upgrading such an application from env_logger 0.4 to env_logger 0.5 that chrono and then later humantime were added as dependencies. In our situation this is a regression, since it increases the number of dependencies (and thus the application size and the build time) compared to env_logger 0.4. We can't stay with env_logger 0.4 because some other libraries we use depend on env_logger 0.5 already; we're actually trying to optimize the build so that only only version of env_logger (0.5) needs to be built.

For these reasons, it would be beneficial to make the dependency on humantime, and the default timestamping functionality in general, optional, e.g. by making them configurable via a default feature, perhaps named "timestamp".

KodrAus:
Related to #71

Thanks for the input @briansmith! I've been wondering about this, but haven't had a strong enough opinion either way to investigate it further.

It seems reasonable for `env_logger` to be built with `--no-default-features` and avoid pulling in dependencies you don't need if that's important to you. The switch from `chrono` to `humantime` should be an improvement in compile times and binary size. I'd rather frame these features as a way to avoid unnecessary dependencies, rather than a way to disable the timestamp in the default format or disable colour.

Perhaps we could have 2 features that control the dependencies we have now:

- `timestamp = ["humantime"]`
- `style = ["termcolor", "atty"]`

The `Formatter` abstraction _should_ encapsulate `termcolor` and `atty` enough to swap them out, besides the check to see if the buffer's style is equal to the thread-local one.

What do you think? Do you want to submit a PR for this?

briansmith:
> The Formatter abstraction should encapsulate termcolor and atty enough to swap them out, besides the check to see if the buffer's style is equal to the thread-local one.

[...]

PR #75, opened 2018-03-10
Wrap termcolor's color api

KodrAus:
Noted in #72 

The `pub use termcolor::Color` we released `env_logger 0.5.0` with was a bad call.

This PR duplicates the `termcolor::Color` API in `env_logger` so breaking changes in `termcolor` can be managed with some gymnastics in `env_logger`.

If a new variant is added to `termcolor::Color`:

- `Color::from_termcolor` will return `None` until/unless we have a corresponding variant
- `Color::FromStr` will return `Err::Unrecognized` until/unless we have a corresponding variant

If `termcolor` drops the concept of `Color` completely then we have a little more work to do, but it should still all be manageable.

This will technically be a breaking change, since our `ParseColorError` isn't equivalent to `termcolor::ParseColorError` so you couldn't `?` it and `Color` isn't equivalent to `termcolor::Color`, but pragmatically I think this is safe to make in a patch.

cc @BurntSushi @briansmith what do you think?

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -575,6 +575,8 @@ impl fmt::Display for Timestamp {
     }
 }
 
+// The `Color` type is copied from https://github.com/BurntSushi/ripgrep/tree/master/termcolor
+
 /// The set of available colors for the terminal foreground/background.
 ///
 /// The `Ansi256` and `Rgb` colors will only output the correct codes when
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -37,14 +37,14 @@
 use std::io::prelude::*;
 use std::{io, fmt};
 use std::rc::Rc;
+use std::str::FromStr;
+use std::error::Error;
 use std::cell::RefCell;
 use std::time::SystemTime;
 
-use termcolor::{ColorSpec, ColorChoice, Buffer, BufferWriter, WriteColor};
+use termcolor::{self, ColorSpec, ColorChoice, Buffer, BufferWriter, WriteColor};
 use atty;
 use humantime::format_rfc3339_seconds;
-
-pub use termcolor::Color;
 
 /// A formatter to write logs into.
 ///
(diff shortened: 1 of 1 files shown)


briansmith:
I think this is a positive change.

My main motivation with this kind of change is eliminating the termcolor dependency or at least making it optional. This appears to be progress in achieving that. My more general motivation for this kind of change is eliminating complexity. IMO, rather than have a more complete color API in env_logger to avoid the termcolor dependency, it would be better to have a simpler API such that, if somebody chooses to enable the termcolor dependency, then they get the color API, but if they don't choose that dependency, then they just don't get the color API and all color parsing and whatever is just a no-op. In fact, I wish the API were different so that all the styling were only available through a `Style` API that is only available when termcolor is used, so that all styling code (not just colors) is avoided when no styling is wanted.

However, again, this is a positive change, even if I don't think it is optimal.

KodrAus:
You're right, I think we can do things differently in `env_logger` with more freedom over the API next time we want to make breaking changes to `env_logger`.

This PR pulls in the complete API that's available now through `pub use termcolor::Color`, so we can remove the public dependency in a patch.

I think the next incremental step would be to move all color and styling APIs into a submodule under `fmt` (even if it's just re-exported) so it'll be easier to conditionally exclude it.

briansmith:
> You're right, I think we can do things differently in env_logger with more freedom over the API next time we want to make breaking changes to env_logger.

This is a breaking change already, right? It is possible (perhaps unlikely) that somebody is already relying on the fact that `env_logger::Color` is the same type as `termcolor::Color`.

KodrAus:
It is technically a breaking change. I looked through `env_logger`'s public dependents using `Color` and none of them seem to depend on `env_logger::Color` and `termcolor::Color` being the same (we've never advertised `Color` as coming from `termcolor`). So I'm comfortable pushing it through as a patch with details in the release notes unless folks are deeply opposed, in which case it might be a while before we're ready to bundle it with other breaking changes.

PR #77, opened 2018-04-06
Add note about dashes

lnicola:


lnicola:
@KodrAus Yup, I figured out the reason, but this took me a while to figure because the logs were working on one computer and not on another. The difference was the target (I assume release builds show less), but it was still confusing :smile:.

Do you think the explanation should be mentioned in the docs?

KodrAus:
Thanks @lnicola! I'll merge this in.

PR #80, opened 2018-04-12
Add Timestamp::source

nvzqz:
This allows for custom timestamp formatting.

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -148,6 +148,13 @@ pub struct StyledValue<'a, T> {
 /// [`Formatter`]: struct.Formatter.html
 pub struct Timestamp(SystemTime);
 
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks for the PR @nvzqz! This looks good to me, but I'll just let it bake mentally for a bit and make sure there's no implications to consider. I'm just not sure if it's better to call `let sys_timestamp = buf.timestamp().source()` with this new method to fetch a `SystemTime` rather than `let sys_timestamp = SystemTime::now()` in your custom format.

nvzqz:
Ah, I didn't see that `Formatter::timestamp` just requests the current system time. For some reason I thought it would be stored somewhere. In that case, this is extra unnecessary functionality.

PR #81, opened 2018-04-15
Add support for default env values

### LINKED ISSUE #47: Accept default log level
I noticed this a couple of times now: I'm considering using log and env_logger for my logging, but env_logger would require that I set RUST_LOG every time I want to actually have output, or I wouldn't get anything at all. What I want to do would be something along these lines:

```rust
env_logger::init_or("something=info");
```

So, if `$RUST_LOG` is not set, env_logger would use `something=info` instead. This could be overwritten with `RUST_LOG= ./app`, causing env_logger to use an empty string.

KodrAus:
Closes #47 

# Default Values for Environment Variables

Adds 2 new methods to `Env` that can be used to supply a default value for environment variables if they're not set:

```rust
fn filter_or(name: impl Into<Cow<'a, str>>, default: impl Into<Cow<'a, str>>);
fn write_style_or(name: impl Into<Cow<'a, str>>, default: impl Into<Cow<'a, str>>);
```

The default value is a string with the same format as the values accepted by the environment variables. I thought this offered some simple symmetry between values from the environment and the defaults, even though it's not so type safe.

As an example, we can default the level to `info` if `MY_LOG_LEVEL` isn't set:

```rust
let env = Env::default()
    .filter_or("MY_LOG_LEVEL", "info")
    .write_style("MY_LOG_STYLE");

env_logger::init_from_env(env);
```

This also means you need to specify the name of the environment variable you want to use, but we could export constants for the default ones if folks don't want to write `RUST_LOG` manually.

### DIFF
--- a/examples/default.rs
+++ b/examples/default.rs
@@ -23,8 +23,8 @@ use env_logger::Env;
 
 fn main() {
     let env = Env::default()
-        .filter("MY_LOG_LEVEL")
-        .write_style("MY_LOG_STYLE");
+        .filter_or("MY_LOG_LEVEL", "trace")
+        .write_style_or("MY_LOG_STYLE", "always");
 
     env_logger::init_from_env(env);
 
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -709,7 +709,7 @@ impl FromStr for Color {
 
     fn from_str(s: &str) -> Result<Color, ParseColorError> {
         let tc = termcolor::Color::from_str(s).map_err(ParseColorError::termcolor)?;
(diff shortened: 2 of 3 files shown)


KodrAus:
A potential downside of this is that we're now very much splitting the builder configuration between two different types with different semantics.

The `Env` type lets us specify the environment variables to read from and a default value if they're missing. The `Builder` type lets us specify the format and set the values read from environment variables explicitly (but not provide defaults).

I think this is ok for now, but the next time we want to make breaking changes it should probably be revisited.

PR #82, opened 2018-04-16
Rework the default format

### LINKED ISSUE #74: Tweaks to the default format and format stability
Our current `env_logger` format is pretty good, but I think there are a few tweaks we can make to it that would improve readability, parsability and compactness. Before that though, I think it would be good if we thought about whether or not we consider the default format as part of our semver'd API.

# Changing the default format in patches

I think we should be mindful, but not against changing the default format in patches. We do eventually want it to be fairly stable, but if a user needs a truly stable format then they should use a custom format.

I'm keen to hear what other people think about this!

# Improvements to our default format

I think we can make some tweaks to our current format that should be an improvement for the majority of users:

- Wrap parts of the format that aren't the actual message in brackets so they're easier to distinguish from the message. This also means all logs (unless you disable all optional parts of the format) will start on a new line with the same character.
- Shorten the log level so they're all the same length. So:
  - `Trace => TRC`
  - `Debug => DBG`
  - `Info => INF`
  - `Warn => WRN`
  - `Error => ERR`

To see how this might look, here's a few examples using our current format, and the changes I was thinking about:

Proposed format:
proposed-all

Current format:
current-all

Proposed format:
proposed-multi-line

Current format:
current-multi-line

This is also a bit subjective, so I'd like to know what you all think about the current and proposed format too!

KodrAus:
Closes #74 

See #74 for more details.

This PR tweaks the default format so it's more readable, parsable and sortable. It also adds a convenient `level_style` method for getting a styled log level that can be displayed. This is still a work in progress. We need some more details about the format stability in the docs.


KodrAus:
The samples in the readme and docs of log output all need to be updated.

KodrAus:
I've also thrown in something to fix #107, so that compiling with `default-features = false` will use standard library macros that do support the special test capturing.

README.md
```diff
@@ -138,3 +138,9 @@ if env::var("RUST_LOG").is_ok() {
 }
 builder.init();
 ```
+
+## Stability of the default format
+
+The default format won't optimise for long-term stability, and explicitly makes no guarantees about the stability of its output across major, minor or patch version bumps during `0.x`.
+
+If you want to capture or innterpret the output of `env_logger` programmatically then you should use a custom format.
```
KodrAus: Typo

src/fmt/mod.rs
```diff
@@ -51,151 +50,202 @@ pub(super) mod pub_use_in_super {
...
-/// builder.format(|buf, record| writeln!(buf, "{}: {}", record.level(), record.args()));
-/// ```
-///
-/// [`Write`]: https://doc.rust-lang.org/stable/std/io/trait.Write.html
-/// [`writeln`]: https://doc.rust-lang.org/stable/std/macro.writeln.html
-/// [`style`]: #method.style
-pub struct Formatter {
-    buf: Rc<RefCell<Buffer>>,
-    write_style: WriteStyle,
+pub(super) struct Format {
```
KodrAus: Call this `FormatBuilder`?

PR #90, opened 2018-07-19
deps: update termcolor to 1

ignatenkobrain:


KodrAus:
Thanks @ignatenkobrain! The build failure is unrelated (looks like we need to bump up our minimum `rustc` version) so I'll merge this in and fix the min version build.

PR #96, opened 2018-07-23
add some docs on defaulting env vars

### LINKED ISSUE #93: Support and clearly document a means to change the default log level
I often write projects in Python where my logging level should be `warn` or even `info` if the user doesn't specify anything.

Currently, the docs are unclear whether there's a way to set a logging level, but still allow it to be overridden by `RUST_LOG`, other than by checking for the `RUST_LOG` environment variable and setting it if it's absent before calling `env_logger::init()`... a solution which, I think you'll agree, looks like an ugly hack.

KodrAus:
Closes #93 

Adds a section to the crate root docs on defaulting environment variables if they're missing.

cc @ssokolow How does this look to you?

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -177,8 +177,30 @@
 //! }
 //! ```
 //! 
+//! ## Specifying defaults for environment variables
+//! 
+//! `env_logger` can read configuration from environment variables.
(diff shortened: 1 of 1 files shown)


ssokolow:
Looks good to me.

PR #99, opened 2018-08-13
Move default level style to a public function

upsuper:
This allows users to write their custom format functions more easily.

### DIFF
--- a/src/fmt.rs
+++ b/src/fmt.rs
@@ -42,6 +42,7 @@ use std::error::Error;
 use std::cell::RefCell;
 use std::time::SystemTime;
 
+use log::Level;
 use termcolor::{self, ColorSpec, ColorChoice, Buffer, BufferWriter, WriteColor};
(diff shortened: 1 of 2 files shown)


KodrAus:
Thanks for the PR @upsuper! This is definitely something I think we should do. I played with it on another branch and ended up making the `StyledValue` take a `Cow<'a, T>` instead of just a `&'a T`. That way, we could return an owned `StyledValue` from methods, so we could capture the log level and avoid having to specify it once to get the style and then again to print it.

Having said that though, I think it's convenient to be able to match the style of the level if you want to apply it to other items, so I think the method you've written is a great addition as-is, and we can follow up with a `styled_value` method that returns the printable value for a given level.

upsuper:
Yeah, I thought about that as well, but I think having an orthogonal function for style should be useful as itself. But having another returning the styled value would be a convenient addition.

PR #100, opened 2018-08-14
Make all dependencies optional

### LINKED ISSUE #71: Allow disabling of the termcolor and atty dependencies
Some environments will never, or will practically never, output to a terminal, or otherwise do not ever need or want stylized output. This is often the case for an application that is intended to be deployed in a containerized environment like Kubernetes or Docker, for example.

Because Cargo downloads each dependency serially, and because Cargo builds each dependency serially, every single dependency inhibits the build performance by some constant factor that's independent of the complexity of the dependency. Further, some applications have continuous integration and/or container build systems that have to rebuild everything from scratch any time any dependency changes, and/or on every commit.

I noticed when upgrading such an application from env_logger 0.4 to env_logger 0.5 that `termcolor` and `atty` were added as dependencies. In our situation this is a regression, since it increases the number of dependencies (and thus the application size and the build time) compared to env_logger 0.4. We can't stay with env_logger 0.4 because some other libraries we use depend on env_logger 0.5 already; we're actually trying to optimize the build so that only only version of env_logger (0.5) needs to be built.

For these reasons, it would be beneficial to make the dependencies on termcolor and atty optional, controlled by a default feature, perhaps named "style" or "tty", so that we can use `default-features = false` to disable them. Or, otherwise, provide a way to get the configurability by environment variables that env_logger provides, without these dependencies.

### LINKED ISSUE #72: Allow disabling the humantime dependency

KodrAus:
Except for `log`.

Closes #71
Closes #72

This PR refactors the internals of the `fmt` module so we can optionally exclude dependencies and their supported features. The goal of making things optional is to improve compile times and reduce artifact sizes so rather than trying to shim the same public API we exclude swathes of API that are tied to specific dependencies when those dependencies aren't available.

By default all current dependencies are included. Since `env_logger` usually lives at the top of a dependency graph, if an application disables default features then it should be unlikely that another dependency adds them back in.

This is still in a bit of a strawman phase, there might be other approaches we'll want to consider.

### DIFF
--- a/src/fmt/termcolor/shim_impl.rs
+++ b/src/fmt/termcolor/shim_impl.rs
@@ -1,43 +1,53 @@
-use std::fmt;
-use std::io;
+use std::io::{self, Write};
 
-use ::WriteStyle;
+use fmt::{WriteStyle, Target};
 
 pub(in ::fmt) mod pub_use_in_super {
     
 }
 
-pub(in ::fmt) struct BufferWriter {}
-pub(in ::fmt) struct Buffer {}
+pub(in ::fmt) struct BufferWriter {
+    target: Target,
+}
+
+pub(in ::fmt) struct Buffer(Vec<u8>);
 
 impl BufferWriter {
-    pub(in ::fmt) fn stderr(color_choice: WriteStyle) -> Self {
-        unimplemented!()
+    pub(in ::fmt) fn stderr(_: WriteStyle) -> Self {
+        BufferWriter {
+            target: Target::Stderr,
+        }
     }
 
-    pub(in ::fmt) fn stdout(color_choice: WriteStyle) -> Self {
-        unimplemented!()
+    pub(in ::fmt) fn stdout(_: WriteStyle) -> Self {
+        BufferWriter {
+            target: Target::Stdout,
+        }
     }
 
     pub(in ::fmt) fn buffer(&self) -> Buffer {
-        unimplemented!()
+        Buffer(Vec::new())
     }
 
     pub(in ::fmt) fn print(&self, buf: &Buffer) -> io::Result<()> {
-        unimplemented!()
+        match self.target {
+            Target::Stderr => io::stderr().write_all(&buf.0),
+            Target::Stdout => io::stdout().write_all(&buf.0),
+        }
     }
 }
 
 impl Buffer {
     pub(in ::fmt) fn clear(&mut self) {
-        unimplemented!()
+        self.0.clear();
     }
 
     pub(in ::fmt) fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
-        unimplemented!()
+        self.0.extend(buf);
+        Ok(buf.len())
     }
 
     pub(in ::fmt) fn flush(&mut self) -> io::Result<()> {
-        unimplemented!()
+        Ok(())
     }
 }
--- a/src/fmt/termcolor/shim_impl.rs
+++ b/src/fmt/termcolor/shim_impl.rs
@@ -3,7 +3,7 @@ use std::io;
 
 use ::WriteStyle;
 
-pub(in ::fmt) mod pub_use {
+pub(in ::fmt) mod pub_use_in_super {
     
 }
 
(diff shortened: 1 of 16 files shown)


KodrAus:
A potential source of breakage here is if folks are currently using `no_default_features` with `env_logger` to exclude `regex`. With this change that will now also exclude styling and timestamps.

KodrAus:
Possibly also worth noting that libraries using `env_logger` as a `dev-dependency` won't cause the default features to be activated if you happen to depend on them.

KodrAus:
cc @briansmith

This makes a pretty big impact on compile-times. On my machine locally for release builds I'm waiting ~18s with default features, and ~1.5s without them. Artifact size is also a bit smaller.

I'll do a proper review of the APIs that are available and unavailable depending on features, but this seems worth following through for folks that care about compile times and don't care about the format.

What do you think?

fbernier:
Thank you for this. The community needs more of these PRs.

src/fmt/termcolor/shim_impl.rs
```diff
@@ -0,0 +1,53 @@
...
+            target: Target::Stdout,
+        }
+    }
+
+    pub(in ::fmt) fn buffer(&self) -> Buffer {
+        Buffer(Vec::new())
+    }
+
+    pub(in ::fmt) fn print(&self, buf: &Buffer) -> io::Result<()> {
+        match self.target {
```
KodrAus: These should probably acquire a lock on the stream first, since it might make multiple calls to `write`.

KodrAus:
I did a quick run through source on GitHub and came up with a few dependencies that would see a change in their default format following this PR:

- [`tower-h2`](https://github.com/tower-rs/tower-h2/blob/1299be66fee7be919699d7c1edfb30adac03d4c1/Cargo.toml#L30)
- [`mesabox`](https://github.com/mesalock-linux/mesabox/blob/7ad81d5b3a0017dea304a9b20eef8e41cafd0973/Cargo.toml#L102)
- [`compiletest`](https://github.com/rust-lang/rust/blob/fa5b3d31fab43d904a7dc6bfe7db5f0e5fb81c43/src/tools/compiletest/Cargo.toml#L8) (and other tools within `rust-lang`)
- [`srt-transport`](https://github.com/russelltg/srt-rs/blob/06ccf6ad40aa27c53209f27d932acd6220141a7d/Cargo.toml#L23)
- [`quickcheck`](https://github.com/BurntSushi/quickcheck/blob/9b72d53e446d3b4941cd382182060e6f8f782e44/Cargo.toml#L25)

I'm starting to lean towards pushing this and #82 together as a minor version bump, so `0.6.0` where we can start that version with clear semantics about how changes to the default format will be handled. I'll open up an issue to get some feedback.

KodrAus:
I've added a simple test harness as a local crate that runs tests for all feature permutations. The number of permutations will increase exponentially with the number of optional features so in the future we might need a different strategy for maintaining confidence that everything plays nicely.

PR #101, opened 2018-08-24
update some debs for minimal-versions

Eh2406:
This bumps the minimal acceptable versions in Cargo.toml to versions that are compatible with `-Z minimal-versions`. This is part of the process of seeing how hard this is for crates to use in preparation for getting it stabilized for use in CI, specifically upstreaming the changes required to get criterion working with it. It is easy to use if all of your dependencies support it, but much harder if trying to impose it on them.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,10 +18,10 @@ status = "actively-developed"
 
 [dependencies]
 log = { version = "0.4", features = ["std"] }
-regex = { version = "1", optional = true }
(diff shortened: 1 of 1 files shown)


KodrAus:
Thanks @Eh2406!

Eh2406:
When this is in a released version I can make a PR to `handlebars`. When that is released I can get `criterion` working with `minimal-versions`. The end is in sight!

PR #126, opened 2019-01-26
filter: Don't print to stdout the warnings due to an invalid spec.

emilio:
This makes programs that rely on the output of other programs not work properly
in these cases. stderr is more appropriate for such things.

Indeed, if you `export RUST_LOG="xxx=xxx"` before building this crate for the
first time, the build fails :)

See https://github.com/rust-lang/rust-bindgen/issues/1502.


emilio:
Thank you!

PR #134, opened 2019-05-31
Indentation for multiline log messages

95ulisse:
The default formatter does not handle multiline messages at all: if a log entry contains a newline,
the second line gets behind the header, which can appear a bit strange.

This PR implements 4 different strategies to handle them, represented by the new enum `fmt::Indent`:

- `Indent::None`: this is the current and default behaviour.
- `Indent::Spaces(usize)`: indent next lines of a fixed amount of spaces.
- `Indent::Auto`: indent the next lines so that they are all aligned to the header of the first line.
- `Indent::RepeatHeader`: repeat the header for each line.

The idea is to expose the indentation as a setting of the default formatter (by adding a new function `default_format_indent` to the `Builder`), so that the user can choose his preferred style. The default setting is obviously `Indent::None` for backward compatibility.

Thank you for your time!

# Example

This is a screenshot of running the following program with different indentation settings in the following order: `None` (the default), `Spaces(4)`, `Auto`, `RepeatHeader`.

```rust
#[macro_use]
extern crate log;
extern crate env_logger;

use log::LevelFilter;
use env_logger::fmt::Indent;

fn main() {
    env_logger::Builder::from_default_env()
        .filter_level(LevelFilter::Info)
        .default_format_indent(Indent::Auto)
        .init();

    info!("First line\nSecond line\nThird line");
}
```

Indent test

### DIFF
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -310,8 +310,7 @@ impl<'a> DefaultFormat<'a> {
                         let mut first = true;
                         for chunk in buf.split(|&x| x == b'\n') {
                             if !first {
-                                let bar = self.fmt.subtle_style("|");
-                                write!(self.fmt.buf, "\n{:width$}{} ", "", bar, width = self.indent_count)?;
+                                write!(self.fmt.buf, "\n{:width$}", "", width = self.indent_count)?;
                             }
                             self.fmt.buf.write_all(chunk)?;
                             first = false;
@@ -426,7 +425,7 @@ mod tests {
             buf: &mut f,
         });
 
-        assert_eq!("[INFO  test::path] log\n    | message\n", written);
+        assert_eq!("[INFO  test::path] log\n    message\n", written);
     }
 
     #[test]
@@ -447,7 +446,7 @@ mod tests {
             buf: &mut f,
         });
 
-        assert_eq!("[INFO  test::path] log\n| message\n", written);
+        assert_eq!("[INFO  test::path] log\nmessage\n", written);
     }
 
     #[test]
@@ -468,6 +467,6 @@ mod tests {
             buf: &mut f,
         });
 
-        assert_eq!("log\n    | message\n", written);
+        assert_eq!("log\n    message\n", written);
     }
 }
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -356,7 +356,7 @@ impl<'a> DefaultFormat<'a> {
                             if !first {
                                 self.fmt.buf.write_all(&[ b'\n' ])?;
                                 match self.indent_count {
-                                    Some(count) => write!(self.fmt.buf, "{:width$}", "", width = count)?,
+                                    Some(count) => write!(self.fmt.buf, "{:width$}{} ", "", self.fmt.subtle_style("|"), width = count)?,
                                     None => self.fmt.write_header(self.record)?
                                 }
                             }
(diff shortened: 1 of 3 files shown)


KodrAus:
Thanks for the PR @95ulisse!

This is something that would be good to think about, but most multi-line messages I've encountered are errors containing a backtrace, which typically follow their own indentation. Is there a different case you're working with?

Applying our own multiline formatting may become ragged for terminal windows that can't fit the entire message on a single line so end up pushing the overflow to the start of a new line without indentation anyway.

Of the indentation options here I think the most useful is to give subsequent lines a small fixed indentation (like your `Indent::Spaces`). I think pushing all lines to the edge of the header is more likely to result in the terminal reflowing text, and duplicating the header looks like there are multiple logs instead of just one.

95ulisse:
Hi KodrAus, thanks for the reply!

> This is something that would be good to think about, but most multi-line messages I've encountered are errors containing a backtrace, which typically follow their own indentation. Is there a different case you're working with?

Indeed, error messages with backtraces are such a case. Another example might be pretty-printing `Debug` objects with `debug!("{:#?}", ...)`.

An example of multiline messages without their own indentation might be printing an error from [`error-chain`](https://docs.rs/error-chain), which produces just a sequence of lines looking like *"Error: ... / Caused by: ... / Caused by: ..."* and it might be nice to have them aligned.

> Applying our own multiline formatting may become ragged for terminal windows that can't fit the entire message on a single line so end up pushing the overflow to the start of a new line without indentation anyway.

A solution to this might be introducing a maximum number of columns after which a newline is automatically inserted. The limit could also be automatically computed using the width of the terminal window (while I'm certain this is possible on Linux, I'm not entirely sure about Windows). When the output is redirected to a file, though, overflow is not a problem and the logs can be indented how much the user prefers.

Duplicating the header makes a single message appear as multiple ones, but on the other side, it helps parsing the logs: for example, `grep`ping for an error in a log will yield full error information instead of just the first line.

In the end, each choice of indentation style has its pros and cons, and whether they make sense or not depends on the target application. That's why I think leaving the choice to the user is the best way to go.

KodrAus:
Hey @95ulisse, thanks for your patience here. Just trying to think of ways we can simplify our own implementation, I don't think we necessarily need to support every use-case in the default format. So I would definitely be on board with just the number-of-spaces based indentation approach and apply it by default.

95ulisse:
Ok, let's go for the spaces-only solution, then. How many spaces should the default indentation be? `4` could be a good compromise for me.

Also, do you want me to keep pushing on this PR, or should I open a new one?

KodrAus:
Thanks for your patience on this one @95ulisse. I think 2 or 4 spaces should be alright, and if you'd like to keep working on this PR that's totally fine with me!

95ulisse:
Ok, I removed all the other possible options and left only the spaces one. The default is set to 4 spaces.

src/fmt/mod.rs
```diff
@@ -287,7 +291,55 @@ impl<'a> DefaultFormat<'a> {
...
+                    fmt: &'a mut DefaultFormat<'b>,
+                    indent_count: usize
+                }
+
+                impl<'a, 'b> Write for IndentWrapper<'a, 'b>  {
+                    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+                        let mut first = true;
+                        for chunk in buf.split(|&x| x == b'\n') {
+                            if !first {
+                                let bar = self.fmt.subtle_style("|");
```
KodrAus: Hmm, I think for our fixed-width indentation we could get away without the `|` here. It would also simplify any regex-style text extraction that wants to de-indent the message.

95ulisse: Yes, it makes sense, we can also go without the `|`.

KodrAus:
Thanks @95ulisse!

This looks good to me.

PR #137, opened 2019-07-02
Add support for Key-Value data in log records.

tmccombs:
See https://github.com/rust-lang-nursery/log/issues/328

### DIFF
--- a/src/fmt/kv.rs
+++ b/src/fmt/kv.rs
@@ -1,6 +1,10 @@
 use std::io::{self, Write};
 
-use super::Formatter;
+#[cfg(feature = "color")]
+use super::WriteStyle;
+use super::{Formatter, StyledValue};
+#[cfg(feature = "color")]
+use anstyle::Style;
 use log::kv::{source::Source, Error, Key, Value, Visitor};
 
 /// Format function for serializing key/value pairs
@@ -41,7 +45,27 @@ impl<'a, 'kvs> Visitor<'kvs> for DefaultVisitor<'a> {
     fn visit_pair(&mut self, key: Key, value: Value<'kvs>) -> Result<(), Error> {
         // TODO: add styling
         // tracing-subscriber uses italic for the key and dimmed for the =
-        write!(self.0, " {}={}", key, value)?;
+        write!(self.0, " {}={}", self.style_key(key), value)?;
         Ok(())
+    }
+}
+
+impl DefaultVisitor<'_> {
+    fn style_key<'k>(&self, text: Key<'k>) -> StyledValue<Key<'k>> {
+        #[cfg(feature = "color")]
+        {
+            StyledValue {
+                style: if self.0.write_style == WriteStyle::Never {
+                    Style::new()
+                } else {
+                    Style::new().italic()
+                },
+                value: text,
+            }
+        }
+        #[cfg(not(feature = "color"))]
+        {
+            text
+        }
     }
 }
--- a/src/fmt/kv.rs
+++ b/src/fmt/kv.rs
@@ -1,0 +1,47 @@
+use std::io::{self, Write};
+
+use super::Formatter;
+use log::kv::{source::Source, Error, Key, Value, Visitor};
+
+/// Format function for serializing key/value pairs
+///
+/// This function determines how key/value pairs for structured logs are serialized within the default
+/// format.
+pub(crate) type KvFormatFn = dyn Fn(&mut Formatter, &dyn Source) -> io::Result<()> + Sync + Send;
+
+/// Null Key Value Format
+///
+/// This function is intended to be passed to
+/// [`Builder::format_key_values`](crate::Builder::format_key_values).
+///
+/// This key value format simply ignores any key/value fields and doesn't include them in the
+/// output.
+pub fn hidden_kv_format(_formatter: &mut Formatter, _fields: &dyn Source) -> io::Result<()> {
+    Ok(())
+}
+
+/// Default Key Value Format
+///
+/// This function is intended to be passed to
+/// [`Builder::format_key_values`](crate::Builder::format_key_values).
+///
+/// This is the default key/value format. Which uses an "=" as the separator between the key and
+/// value and a " " between each pair.
+///
+/// For example: `ip=127.0.0.1 port=123456 path=/example`
+pub fn default_kv_format(formatter: &mut Formatter, fields: &dyn Source) -> io::Result<()> {
+    fields
+        .visit(&mut DefaultVisitor(formatter))
+        .map_err(|e| io::Error::new(io::ErrorKind::Other, e))
+}
+
+struct DefaultVisitor<'a>(&'a mut Formatter);
+
+impl<'a, 'kvs> Visitor<'kvs> for DefaultVisitor<'a> {
+    fn visit_pair(&mut self, key: Key, value: Value<'kvs>) -> Result<(), Error> {
+        // TODO: add styling
+        // tracing-subscriber uses italic for the key and dimmed for the =
+        write!(self.0, " {}={}", key, value)?;
+        Ok(())
+    }
+}
--- a/src/fmt/kv.rs
+++ b/src/fmt/kv.rs
@@ -5,7 +5,7 @@ use super::WriteStyle;
 use super::{Formatter, StyledValue};
 #[cfg(feature = "color")]
 use anstyle::Style;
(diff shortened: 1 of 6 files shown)


src/fmt/mod.rs
```diff
@@ -289,6 +294,28 @@ impl<'a> DefaultFormat<'a> {
     fn write_args(&mut self, record: &Record) -> io::Result<()> {
         writeln!(self.buf, "{}", record.args())
     }
+
+    fn write_kv(&mut self, record: &Record) -> io::Result<()> {
+        if !self.key_values {
+            return Ok(())
+        }
+        let kvs = record.key_values();
+        if !self.written_header_value && kvs.count() > 0 {
```
tmccombs: I don't love this, but I couldn't think of a way of doing it I liked any better.

src/fmt/mod.rs
```diff
@@ -289,6 +294,28 @@ impl<'a> DefaultFormat<'a> {
...
+            self.written_header_value = true;
+            let open_brace = self.subtle_style("[");
+            write!(self.buf, "{}", open_brace)?;
+        }
+        // Ideally we would be able to transport the original io::Error through the visit
```
tmccombs: See https://github.com/rust-lang-nursery/log/pull/336

simao:
Any news on this? Could we at least merge it behind a feature flag or something? Does this need any changes besides a rebase?

hendrikmaus:
I would also love to start using it.

One good addition to the pull-request would probably be a usage example.

simao:
I tried these changes and they work well. As far as usage examples go, I think that is done in the `log` crate? https://docs.rs/log/0.4.16/log/#structured-logging

Users of env_log would just need to use the `log/kv_unstable` feature, which I suppose should be mentioned somewhere?

One thing I noticed is maybe the target should be written after the kv pairs. Currently we get:

```
[2022-04-01T08:52:43Z INFO  tide::server] Server listening on http://0.0.0.0:8080
[2022-04-01T08:52:50Z INFO  method=POST path=/sensors/inside_weather status=200 - OK duration=11.487824ms tide::log::middleware] --> Response sent
```

But IMO the following looks cleaner, since target stays aligned with lines that do not have kv pairs:

```
[2022-04-01T08:54:12Z INFO  tide::log::middleware method=POST path=/sensors/inside_weather status=200 - OK duration=10.166265ms] --> Response sent
```

Also wanted to mentioned that it might be worth it to consider putting the structured data in different lines, to improve readability. At least for development, IMO the following looks better:

```
[2022-04-01T08:57:16Z INFO  tide::server] Server listening on http://0.0.0.0:8080
[2022-04-01T08:57:19Z INFO  tide::log::middleware
    method POST
    path /sensors/inside_weather] <-- Request received
[2022-04-01T08:57:19Z INFO  tide::log::middleware
    method POST
    path /sensors/inside_weather
    status 200 - OK
    duration 10.842409ms] --> Response sent
```

This is what `femme` does, and it's just using `write!(self.0, "\n    {} {}", key, value)` in `visit_pairs`. 

Though in production we'd probably want everything on the same line.

hendrikmaus:
Thanks for pointing to the examples in the log crate. That makes sense. 

As to your other points: I'd second both of them.

svanharmelen:
I was just about to open an issue when I found this PR... It would be awesome if this could be merged so I can keep using `env_logger` and start using structured logging!

svanharmelen:
Thanks for the rebase @tmccombs! After using this a little I'm thinking that it would (maybe) be good to make it configurable where the K/V's are written? As I'm used that the are written behind the actual log message instead of in front of it.

When reading a log of log files that all log a different number of K/V's, you really have to scan al lot for where the actual message is. Where if the K/V's are added at the end, all the messages are directly above each other 🤷🏻‍♂️

tmccombs:
I added a commit that moves the Key values behind the rest of the option, and made it configurable whether it is on a single line or on separate lines, as @simao suggested.

svanharmelen:
Looks really good!! Thanks for your efforts, I'll be using your branch until this get (hopefully) merged 👍🏻

fgimian:
I suggest also checking out frameworks like [zerolog](https://github.com/rs/zerolog) which present structured log fields as `key=value` fields on the same line and with pretty colours.

image

I think output similar to this would be nicer personally 😄

prasannavl:
Any reason on what this is blocked on? Would be great to get this merged, even until an experimental flag.

epage:
@tmccombs at this point my thoughts are
- Mirror what tracing-subscriber does for formatting and control of it
- Rename the feature to `unstable-kv`, other features are using dashes at this point and I tend to put `unstable-` first.

epage:
Hadn't looked into that before.  Thanks for listing those considerations.  As this is unstable, let's focus on an MVP and just use the formatting they do by default and not provide any configuration.  We can then explore in an Issue what our formatting configuration options are, both those from the two PRs and from tracing-subscriber.

src/fmt/kv.rs
```diff
@@ -0,0 +1,45 @@
+use std::io::{self, Write};
+
+use super::Formatter;
+use log::kv::{source::Source, Error, Key, Value, Visitor};
+
+/// Format function for serializing key/value pairs
+///
+/// This trait determines how key/value pairs for structured logs are serialized within the default
+/// format.
+pub(crate) type KvFormatFn = dyn Fn(&mut Formatter, &dyn Source) -> io::Result<()> + Sync + Send;
```
epage, line 10: Is there a reason you didn't match `FormatFn` in making this `Box`ed?

epage:
`KvFormatFn` also works and fits within the style of the existing API.  Thanks!

src/fmt/kv.rs
tmccombs: Because I use it as a `&` [here](https://github.com/rust-cli/env_logger/pull/137/files#diff-3d324710ba49d50da6915fa5a9bee1c6fbd2a8a97a7d8aa45730c3f9a71f54baR284). But if you would prefer fo rme to make it boxed, I'm fine with that too.

tmccombs:
I've resolved conflicts and squashed the commits.

src/fmt/kv.rs
epage: Prefer consistency but not strong enough

src/fmt/mod.rs
```diff
@@ -70,6 +91,38 @@ pub enum TimestampPrecision {
     Nanos,
 }
 
+/// Style for displaying key/value structured data
+///
+/// This determines how key/value pairs are displayed in the log output.
+/// The values can be suppressed, included in a trailer at the end of the line, or
+/// included as multiple lines after the message.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum KVStyle {
```
epage, line 100: Is this a zombie from resolving the merge conflicts?

src/fmt/mod.rs
```diff
@@ -41,13 +58,17 @@ use log::Record;
...
+mod kv;
 pub(crate) mod writer;
 
 #[cfg(feature = "color")]
 pub use anstyle as style;
 
 #[cfg(feature = "humantime")]
 pub use self::humantime::Timestamp;
+#[cfg(feature = "unstable-kv")]
+pub use self::kv::*;
```
epage, line 71: If you are already going back to make a change, can you you check this on `RUSTDOCFLAGS=--cfg=docsrs cargo +nightly doc --all-features` (I'm hoping that is correct).

Each API item behind a `cfg` should have a note in the documentation, like [`clap::command` requiring `cargo`](https://docs.rs/clap/latest/clap/macro.command.html).

Most likely, you'll need to either duplicate this `cfg` on the items or move the `cfg` into the `kv` mod.

src/fmt/mod.rs
tmccombs: ah, yes it is

src/fmt/mod.rs
tmccombs: Looks like it works as is:

image

image

KodrAus:
Hey all! :wave:

Firstly, thank you for picking up and carrying this project forwards 🙏

We're (finally) about to stabilize the key-value support in `log` (see https://github.com/rust-lang/log/pull/617), making this possible to support in `env_logger` without needing to fiddle with any unstable feature flags. If you'd like any help finishing off support here please do let me know.

tmccombs:
I've updated to use the stable version of the log kv feature. Fixed merge conflicts, and fixed the compilation error when the color option is not enabled.

PR #140, opened 2019-09-09
Add more timestamp precisions

### LINKED ISSUE #122: timestamps with milliseconds
I read through https://github.com/sebasmagri/env_logger/issues/87 and I can't help but (humbly) second-guess (no pun intended) the decision. In my opinion, second-level precision is too coarse, and nanosecond-level is too fine, and milliseconds are the happy medium. And I would argue that it doesn't even need to be configurable. I'd propose that if `default_format_timestamp` logged millisecond-level timestamps, that would be good enough for almost everybody. I realize `humantime` doesn't support milliseconds, so I guess this would mean going back to chronos, unless we can convince humantime to add that...

oherrala:
This PR adds two more timestamp precisions: milliseconds and microseconds. 

I sprinkled some `#[deprecated]` tags around for things I consider could be removed in the future since the same functionality can be achieved in other way.

This probably still have sharp corners to work on.

Fixes #122

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ members = [
 log = { version = "0.4", features = ["std"] }
 regex = { version = "1.0.3", optional = true }
 termcolor = { version = "1.0.2", optional = true }
-humantime = { version = "1.1", optional = true }
+humantime = { version = "1.3", optional = true }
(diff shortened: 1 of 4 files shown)


oherrala:
@KodrAus Rebase done. Travis seems happy.

PR #142, opened 2019-09-23
Update to 2018 edition

KodrAus:
Updates to the 2018 edition, removes deprecated items, and run `cargo fmt` to tidy things up a bit.


KodrAus:
This is a breaking change, so we'd need to release it as `0.7.0`.

Issue #145, opened 2019-10-16
Panic due to destroyed TLS value


divergentdave:
I was running a multithreaded program with lots of logging statements under GDB, and I got the following panic due to a log statement coming from inside a Drop implementation. (The drop calls were invoked due to crossbeam's epoch-based GC)

`panicked at 'cannot access a TLS value during or after it is destroyed: AccessError'`

```
  10: core::result::unwrap_failed               
             at src/libcore/result.rs:1084                           
  11: core::result::Result<T,E>::expect                               
             at /rustc/f7af19c279b8b7ea3d2c21fcbd67164af8d5d968/src/libcore/result.rs:879
  12: std::thread::local::LocalKey<T>::with
             at /rustc/f7af19c279b8b7ea3d2c21fcbd67164af8d5d968/src/libstd/thread/local.rs:239
  13: <env_logger::Logger as log::Log>::log
             at /home/david/.cargo/registry/src/github.com-1ecc6299db9ec823/env_logger-0.7.0/src/lib.rs:810
  14: log::__private_api_log
             at /home/david/.cargo/registry/src/github.com-1ecc6299db9ec823/log-0.4.8/src/lib.rs:1393
...
  32: __call_tls_dtors
  33: start_thread
  34: __clone
```

I think the call of `std::thread::LocalKey::with` should be replaced with `try_with`, and in the event of it returning an AccessError, (meaning the TLS key has already been destroyed) a new Formatter should be allocated on the stack, as in the other exceptional code paths.

PR #146, opened 2019-10-16
Thread local fallback

divergentdave:
This PR provides a failing test to reproduce the issue described in #145, and fixes the issue as described by handling the AccessError.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,10 @@ name = "log-in-log"
 harness = false
 
 [[test]]
+name = "log_tls_dtors"
+harness = false
+
+[[test]]
 name = "init-twice-retains-filter"
 harness = false
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -807,7 +807,7 @@ impl Log for Logger {
                 static FORMATTER: RefCell<Option<Formatter>> = RefCell::new(None);
             }
 
-            FORMATTER.with(|tl_buf| {
+            let tls_result = FORMATTER.try_with(|tl_buf| {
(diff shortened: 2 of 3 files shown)


KodrAus:
Thanks for the PR @divergentdave!

It looks like we're accumulating a bit of logic in this method so maybe it's time to refactor it a bit into something more manageable. I was thinking we could just inline the write call into the places where we have a `Formatter`. So something like (probably doesn't actually compile):

```rust
let print = |formatter, record| {
    let _ = (self.format)(&mut formatter, record)
        .and_then(|_| formatter.print(&self.writer));
    formatter.clear();
}

let printed = FORMATTER.try_with(|tl_buf| {
    match tl_buf.try_borrow_mut() {
        // There are no active borrows of the buffer
        Ok(ref mut tl_buf) => match tl_buf {
            // We have a previously set formatter
            Some(ref mut formatter) => {
                if formatter.write_style() != self.writer.write_style() {
                    *formatter = Formatter::new(&self.writer)
                }

                print(formatter, record);
            },
            // We don't have a previously set formatter
            None => {
                let mut formatter = Formatter::new(&self.writer);
                print(formatter, record);

                *tl_buf = Some(formatter);
            }
        }
        // There's already an active borrow of the buffer
        Err(_) => {
            print(&mut Formatter::new(&self.writer), record);
        }
    }
}).is_ok();

if !printed {
    print(&mut Formatter::new(&self.writer), record);
}
```

What do you think?

divergentdave:
Sounds good, I'll split that out

Issue #102, opened 2018-08-25
stdweb


aep:
would it be reasonable to add wasm support via stdweb?

KodrAus:
Hi @aep, I think we should be able to make this work using the `Console` API, and am on-board with supporting it.

It should be easier to manage an implementation specifically for wasm once #100 lands.

huangjj27:
What about using `web-sys` like [scottjmaddox/rust-web-console-logger](https://github.com/scottjmaddox/rust-web-console-logger) or [blm768/wasm-bindgen-console-logger](https://github.com/blm768/wasm-bindgen-console-logger)?

KodrAus:
@huangjj27 that would also work :+1: If anybody would like to take a shot at wasm support that would be amazing 🙂

PR #148, opened 2019-12-17
Add basic wasm32-unknown-unknown support

huangjj27:
Simple version of web logger support

## Update 2020-01-10 23:31:15
This PR is to add basic, ugly but work support for the `wasm32-unknown-unknown` target, with default features disable.  Tests can be used as:

```sh
wasm-pack test --headless --firefox -- --no-default-features --test web
wasm-pack test --node -- --no-default-features --test node
```

## Issues remain
- To separate color, style(maybe still  parts of `fmt`) and writer
- To make wasm support independent to other features

## compile

modified example `default.rs`:
```

#[macro_use]
extern crate log;

use env_logger::Env;

#[cfg(target="wasm32-unknown-unknown")]
use wasm_bindgen::prelude::*;

fn main() {
    // The `Env` lets us tweak what the environment
    // variables to read are and what the default
    // value is if they're missing
    let env = Env::default()
        .filter_or("MY_LOG_LEVEL", "trace")
        .write_style_or("MY_LOG_STYLE", "always");

    env_logger::init_from_env(env);

    trace!("some trace log");
    debug!("some debug log");
    info!("some information log");
    warn!("some warning log");
    error!("some error log");
}

#[cfg(target="wasm32-unknown-unknown")]
#[wasm_bindgen(start)]
fn start() {
    main();
}
```

compile the example and generate bindgen:
```
cargo build --example default --no-default-features
wasm-bindgen ./target/wasm32-unknown-unknown/debug/examples/default.wasm --out-dir=./wasm --web
```

add `index.html` to `./wasm` directory:
```
<!DOCTYPE html>
<html>
  <head>
    <meta http-equiv="Content-type" content="text/html; charset=utf-8"/>
    <title>Tour - Iced</title>
  </head>
  <body>
    <script type="module">
      import init from "./default.js";

      init('./default_bg.wasm');
    </script>
  </body>
</html>
```

open the `index.html` from a browser(I used Firefox 71). It's expected to see logs on the console(F12)


huangjj27:
While implementing this feature, I found the lost of level after formatted, then I have to re parse the level message, Otherwise I have to log all message with console.log. Any better way to get the level message so that we could seperate different levels' message?

huangjj27:
related to #102 #149

Cargo.toml
```diff
@@ -26,6 +26,9 @@ termcolor = { version = "1.0.2", optional = true }
 humantime = { version = "1.3", optional = true }
 atty = { version = "0.2.5", optional = true }
 
+[target.wasm32-unknown-unknown.dependencies]
+wasm-bindgen = "*"
```
KodrAus, line 30:
```suggestion
wasm-bindgen = "0.2.64"
```

examples/default.rs
```diff
@@ -20,6 +20,9 @@ extern crate log;
 
 use env_logger::Env;
 
+#[cfg(all(target_arch = "wasm32", target_vendor = "unknown"))]
```
KodrAus, line 23: Maybe we should add a dedicated example for wasm, even if it's almost exactly the same as this one, just to make it easier to see what's going on.

huangjj27:
@KodrAus Thanks for your sugesstion! I have committed some codes for that.

Another point is that，how could we parse the logging levels on web as we do on native runtime? I think it is better to make use of the logging levels from the web console

huangjj27:
block the merge with  the logging levels issue

jplatte:
Hey @huangjj27, it's a little hard to tell what the current state of this is. Is there more work to be done?

huangjj27:
> Is there more work to be done?

Hi @jplatte , I think there is:
- When I wrote this PR, I hadn't considered how to log with the browsers' logging level, all message is at the `log` level now. I think that integrating with the browser can be improved.
- the CI seems blocked, we need to pass through it

jplatte:
> * When I wrote this PR, I hadn't considered how to log with the browsers' logging level, all message is at the `log` level now. I think that integrating with the browser can be improved.

Do you plan to work on this?

> * the CI seems blocked, we need to pass through it

I'd expect CI to run properly after a rebase.

huangjj27:
> Do you plan to work on this?

@jplatte Hi! I rebased and it passed all checks now. I'm glad that someone will help with this issue. I think it needs much refactoring to the current formatter?

KodrAus:
We could refactor things to pass the `log::Level` all the way through the writer so that `web::print` can pick the right console function to call.

huangjj27:
I think this will pending until refactoring for [this](https://github.com/rust-lang/log/issues/328) is done, for there are information of level and some other meta data we can make use of.

KodrAus:
@huangjj27 Support for structured logging is something we could add at any time, but is absolutely desirable.

I think all we need to do here is add some CI support for WebAssembly. I've got [an example from another project](https://github.com/sval-rs/value-bag/blob/main/.github/workflows/rust.yml#L136-L145), and there's also [the `wasm-bindgen` docs](https://rustwasm.github.io/wasm-bindgen/wasm-bindgen-test/index.html).

If you'd like a hand getting this up and running let me know!

huangjj27:
> If you'd like a hand getting this up and running let me know!

Ok, I will take this at the weekend, or as soon as I’ get spare time

huangjj27:
CI is added but it seems not to test. Will fix it as soon

huangjj27:
Aha! [here](https://rustwasm.github.io/docs/wasm-pack/commands/test.html#running-only-some-tests) is the document.

huangjj27:
encounter with [this bug](https://github.com/rust-lang/rust/issues/48564#issuecomment-505114709).

huangjj27:
@KodrAus  I get it worked, but there is some issue: when I tried to compile with the `wasm32-unknown-unknown` target, it still prevented me from compiling, showing this denied warning:
```log
error: function is never used: `print`
   --> src\fmt\writer\wasm.rs:13:31
    |
13  | pub(in crate::fmt::writer) fn print(msg: &str, t: Target) {
    |                               ^^^^^
    |
note: lint level defined here
   --> src\lib.rs:280:54
    |
280 | #![deny(missing_debug_implementations, missing_docs, warnings)]
    |                                                      ^^^^^^^^
    = note: `#[deny(dead_code)]` implied by `#[deny(warnings)]`

error: aborting due to previous error
```

although the `wasm::print` function is use in `termcolor::shim_impl` :
```
  // termcolor/shim_impl.rs: 32-48
   pub(in crate::fmt::writer) fn print(&self, buf: &Buffer) -> io::Result<()> {
        // This impl uses the `eprint` and `print` macros
        // This is so their output can be captured by `cargo test`
        let log = String::from_utf8_lossy(&buf.0);

        #[cfg(all(target_arch = "wasm32", target_vendor = "unknown"))]
        wasm::print(&log, self.target);

        #[cfg(not(all(target_arch = "wasm32", target_vendor = "unknown")))]
        match self.target {
            Target::Stderr => eprint!("{}", log),
            Target::Stdout => print!("{}", log),
        }

        Ok(())
    }
```

I think there is something I don't know or I mess up so here is the error. If you have some advice please let me know, thanks!

huangjj27:
Hi @KodrAus, basic implement and basic CI are committed. Any Advice?

jplatte:
I've just removed `deny(warnings)` on master, we really shouldn't do that. CI will still fail with rustc or clippy emitting warnings though, since pass `-- -D warnings`. Could it be that (without `deny(warnings)`) the function that `termcolor::shim_impl::print` is also marked unused? IIRC, rustc will complain about transitively dead code too.

sunfishcode:
I'm curious why it's desirable to use env_logger on wasm32-unknown-unknown. The motivation here seems to be for use on the Web, but the Web embedding doesn't have environment variables, and its stdin and stderr don't print anything. The `log` facade is intended to make it easy to plug in different logging backends; would it make sense to write a new backend rather than porting `env_logger`?

huangjj27:
>  would it make sense to write a new backend rather than porting env_logger?

Thanks for advice! I'll abandon this PR.

PR #156, opened 2020-03-11
Change color of DEBUG and TRACE for improved readability

### LINKED ISSUE #138: Debug level is unreadable in light terminal themes
Since Debug level is set to white, it cannot be read in light terminal themes.

https://github.com/sebasmagri/env_logger/blob/8fa62401ceec3a7571565fb3437a7cf4b906f9ee/src/fmt/writer/termcolor/extern_impl.rs#L56

image

ifreund:
This makes the log level much more readable when using a theme with a light background or a low contrast dark theme.

Fixes #138

### DIFF
--- a/src/fmt/writer/termcolor/extern_impl.rs
+++ b/src/fmt/writer/termcolor/extern_impl.rs
@@ -52,7 +52,7 @@ impl Formatter {
         let mut level_style = self.style();
         match level {
             Level::Trace => level_style.set_color(Color::Black).set_intense(true),
-            Level::Debug => level_style.set_color(Color::White),
+            Level::Debug => level_style.set_color(Color::Blue),
             Level::Info => level_style.set_color(Color::Green),
             Level::Warn => level_style.set_color(Color::Yellow),
             Level::Error => level_style.set_color(Color::Red).set_bold(true),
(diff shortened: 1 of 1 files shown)


andrews05:
The blue is apparently totally unreadable on a default Windows terminal: https://github.com/shssoichiro/oxipng/issues/540#issuecomment-1650739328

Can we change this to magenta, perhaps?

ace-dent:
Magenta... 👀 ... perhaps something subdued 🙏 
The colors selected are not great for accessibility...

epage:
So long as we use built-in colors, we leave it to the users terminal theme to ensure the colors are readable.

While `cmd` isn't configurable, `Terminal` is and I'd highly recommend it.

andrews05:
@epage I get that about built-in colours, but magenta is still available so why not use it? Surely it would be better than saying users can't use their system's built-in terminal 🙂
I'd be happy to open a PR if you agree.

epage:
In my mind, the color should stand out proportionate to the severity.  Magenta tends to stand out on all color schemes.  I would more pick that for levels above a red error than for debug or trace.

Also, in the future, please open new issues for concerns like this rather than posting in long closed PRs.  The visibility of a PR conversation is dramatically lower, it isn't in "our backlog", and new PRs aren't an appropriate place for "large" decisions as they also get less visibility and frequently enough multiple PRs get created for a single topic and having PRs focus on anything besides the implementation can spread out the conversation.

PR #164, opened 2020-07-12
move ci to github actions

SirWindfield:
This adds a first draft of a GitHub Actions workflow that does some basic checks on PRs:
- Run rustfmt
- Run clippy
- Run check on all available Rust versions (beta, nightly, stable) + MSRV of 1.31.0
- Run `ci` crate and example

Relies on #165 to successfully run CI.

__Edit:__ I am not sure how to actually test run this, I will probably just create a dummy PR on my fork and report back once they run successfully.

__Edit 2:__ I created a (really ugly) PR on my fork at https://github.com/SirWindfield/env_logger/pull/3/. All checks pass except the lint, which is expected.


SirWindfield:
Ok, so it all works. There is some stuff I want to point out though:

1) Clippy runs against nightly and not 1.31.0, meaning that it picks up lints that are currently not supported on 1.31.0 (like `#[non_exhaustive]`). This should be changed, shouldn't it?
2) Documentation is currently not published to github pages. Not sure if this is actually needed or not. I do think a workflow that listens for tag pushes and automatically publishes to crates.io is more beneficial here.

KodrAus:
Thanks @SirWindfield! This looks great to me. I think for `clippy` running against `nightly` is ok, because we should be able to `#[allow(clippy::lint)]` without producing warnings on channels that we can't implement the lint for. What do you think?

SirWindfield:
Sounds like a good plan. We can fix those later on if the MSRV is bumped up :)

I'd however like to open up an issue for adding a cd workflow as well to make publishing the crate a little bit more automated :)

PR #165, opened 2020-07-12
refactor: fix clippy warnings

SirWindfield:
This fixes all clippy warnings that I encountered

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -312,7 +312,7 @@ fn parse_spec(spec: &str) -> (Vec<Directive>, Option<inner::Filter>) {
     }
     mods.map(|m| {
         for s in m.split(',') {
-            if s.len() == 0 {
+            if s.is_empty() {
                 continue;
             }
             let mut parts = s.split('=');
@@ -354,7 +354,7 @@ fn parse_spec(spec: &str) -> (Vec<Directive>, Option<inner::Filter>) {
         }
     });
 
-    let filter = filter.map_or(None, |filter| match inner::Filter::new(filter) {
+    let filter = filter.and_then(|filter| match inner::Filter::new(filter) {
         Ok(re) => Some(re),
         Err(e) => {
             eprintln!("warning: invalid regex filter - {}", e);
@@ -362,7 +362,7 @@ fn parse_spec(spec: &str) -> (Vec<Directive>, Option<inner::Filter>) {
         }
     });
 
-    return (dirs, filter);
+    (dirs, filter)
 }
 
 // Check whether a level and target are enabled by the set of directives.
--- a/src/fmt/writer/atty.rs
+++ b/src/fmt/writer/atty.rs
@@ -9,8 +9,6 @@ from being printed.
 
 #[cfg(feature = "atty")]
 mod imp {
-    use atty;
-
     pub(in crate::fmt) fn is_stdout() -> bool {
         atty::is(atty::Stream::Stdout)
     }
--- a/src/fmt/writer/termcolor/extern_impl.rs
+++ b/src/fmt/writer/termcolor/extern_impl.rs
@@ -452,6 +452,7 @@ impl_styled_value_fmt!(
 /// Hexadecimal numbers are written with a `0x` prefix.
 #[allow(missing_docs)]
 #[derive(Clone, Debug, Eq, PartialEq)]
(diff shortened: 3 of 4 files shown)


src/fmt/writer/atty.rs
```diff
@@ -9,8 +9,6 @@ from being printed.
 
 #[cfg(feature = "atty")]
 mod imp {
-    use atty;
```
SirWindfield, line 12: unused

src/fmt/writer/termcolor/extern_impl.rs
```diff
@@ -452,6 +452,7 @@ impl_styled_value_fmt!(
 /// Hexadecimal numbers are written with a `0x` prefix.
 #[allow(missing_docs)]
 #[derive(Clone, Debug, Eq, PartialEq)]
+#[non_exhaustive]
```
SirWindfield, line 455: The same as the manual implementation.

src/lib.rs
```diff
@@ -245,6 +245,7 @@
 #![cfg_attr(rustbuild, feature(staged_api, rustc_private))]
 #![cfg_attr(rustbuild, unstable(feature = "rustc_private", issue = "27812"))]
 #![deny(missing_debug_implementations, missing_docs, warnings)]
+#![allow(clippy::needless_doctest_main)]
```
SirWindfield, line 248: The main functions are not needed in doc tests but they make sense here as they reflect real-world usage more accurately.

src/lib.rs
```diff
@@ -260,10 +261,10 @@ use self::fmt::writer::{self, Writer};
 use self::fmt::Formatter;
 
 /// The default name for the environment variable to read filters from.
-pub const DEFAULT_FILTER_ENV: &'static str = "RUST_LOG";
+pub const DEFAULT_FILTER_ENV: &str = "RUST_LOG";
```
SirWindfield, line 264: By default 'static

src/filter/mod.rs
```diff
@@ -354,15 +354,15 @@ fn parse_spec(spec: &str) -> (Vec<Directive>, Option<inner::Filter>) {
         }
     });
 
-    let filter = filter.map_or(None, |filter| match inner::Filter::new(filter) {
+    let filter = filter.and_then(|filter| match inner::Filter::new(filter) {
```
SirWindfield, line 357: `and_then` returns None if the Option contains None, otherwise it returns T. Which is the same as the `map_or` function call.

SirWindfield:
Oh, I didn't know that non_exhaustive is experimental on 1.31. Reverting..

KodrAus:
@SirWindfield `1.31.0` is pretty old now, I'd be up for bumping it to something more recent.

SirWindfield:
Not sure what a sensible good one would be though. Maybe something relative is a good approach here, like supporting the latest four versions, meaning that from today's perspective, Rust `1.44.1` is the newest version, so we'd ensure support for `1.41+`.

KodrAus:
That sounds good to me! I usually bump the MSRV as-needed to something reasonably recent but old enough. As a baseline I check whatever the latest version is that's available in Ubuntu packages for the current LTS (which right now is `1.41.0`).

SirWindfield:
That's a great idea! Sounds good! I can add a `rust-toolchain` file to pin the MSRV.

src/fmt/writer/termcolor/extern_impl.rs
jplatte: This would bump the MSRV from currently 1.31 to 1.40.

src/fmt/writer/termcolor/extern_impl.rs
```diff
@@ -479,6 +480,7 @@ impl Color {
             Color::White => Some(termcolor::Color::White),
             Color::Ansi256(value) => Some(termcolor::Color::Ansi256(value)),
             Color::Rgb(r, g, b) => Some(termcolor::Color::Rgb(r, g, b)),
+            _ => None,
```
jplatte: If we do bump the MSRV, let's also make this function return `Color` instead of `Option<Color>`.

SirWindfield: Great catch, I totally missed that :)

SirWindfield:
By the power that has been given to me, I'll just merge this :)

PR #167, opened 2020-07-18
Rename ci.yml to main.yml

KodrAus:
Just tinkering with our GitHub Actions a bit.


.github/workflows/main.yml
```diff
@@ -1,11 +1,6 @@
 name: Continuous Integration
 
-on:
-  pull_request:
-    paths:
-      - "**.rs"
-      - "Cargo.toml"
-      - "Cargo.lock"
+on: [push, pull_request]
```
SirWindfield, line 3: IMO it doesn't make sense to run it on push. The code is already on the branch (possibly in a broken state if someone pushed directly to master). That's why I set it to PR only and filter out for files that do affect the build. Changing the readme shouldn't trigger the whole run, it's a waste of resources. If you remove the filter, changes to README would still run the whole workflow.

So in my eyes the change doesn't make sense tbh. I would, however, set master as protected so even admins cannot push to it. Meaning we also have to go through PRs. This ensures, that the CI runs even against our code.

SirWindfield:
I would suggest to add a `#[allow(clippy::...)]` to the Color enum in a separate PR to prevent failing CI.

SirWindfield:
On that note I named it CI as it is responsible for the continuous integration of PRs. I am working on a workflow that automatically publishes on crates and github when a new tag is pushed to the repo as well. So the names `cd.yml` and `ci.yml` seemed fitting :)

.github/workflows/main.yml
```diff
@@ -1,11 +1,6 @@
 name: Continuous Integration
 
-on:
-  pull_request:
-    paths:
```
SirWindfield, line 5: I just noticed that the filter should also include the workflow folder as changes to those should be run as well I think... 🤔

.github/workflows/main.yml
KodrAus: @SirWindfield That's fair, I'm not sure we need the file filter though, since we could and possibly should be testing the code fragments in our README, and changes to CI should also run through CI.

Making the main branch protected sounds like a good change though!

SirWindfield: How do you want to test the readme fragments? Afaik, cargo doesn't test those anyway. There has been work on a proof-of-concept, but I wouldn't include that tbh: https://github.com/rust-lang/cargo/issues/383

KodrAus: Ah, there's a library called [`rust-skeptic`](https://github.com/budziq/rust-skeptic) that lets you test readme samples, but it looks like it might not have been updated in a while.

I guess what I'm trying to get at is that it doesn't seem necessary to limit when CI runs based on an includelist glob, since you can come up with reasons to want to run it for just about any filetype, and new filetypes get introduced from time to time.

KodrAus:
This specific changeset doesn't seem like something we'll want to pursue, so I'll go ahead and close it now, and not meddle with our CI until you've had a chance to explore the other features you had in mind @SirWindfield 🙂

.github/workflows/main.yml
SirWindfield: @KodrAus ah, ok now I get it. 

> I guess what I'm trying to get at is that it doesn't seem necessary to limit when CI runs based on an includelist glob, since you can come up with reasons to want to run it for just about any filetype, and new filetypes get introduced from time to time.

Wouldn't an exclude list then make even more sense? that way, new files are added automatically and we have to manually put them into the filter.

PR #168, opened 2020-07-19
Deprecate env_logger::from_env and clean up examples

KodrAus:
The `env_logger::builder` and `env_logger::from_env` methods are a bit confusing as-is, because it's not clear that `env_logger::builder` should call `Builder::from_default_env` instead of `Builder::new`. The `builder` function is a convenient way to get started configuring your logger before initializing it, but the `from_env` method doesn't actually save you any work because you still need to `use env_logger::Env`, so may as well `use env_logger::{Builder, Env}` while you're at it.

I've also cleaned up our examples a bit and added a new one for logging in `cargo test`.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -225,7 +225,7 @@
 //! ```
 //! use env_logger::Env;
 //!
-//! env_logger::from_env(Env::default().default_filter_or("warn")).init();
+//! env_logger::Builder::from_env(Env::default().default_filter_or("warn")).init();
 //! ```
 //!
 //! [log-crate-url]: https://docs.rs/log/
@@ -1185,6 +1185,9 @@ where
 /// Create a new builder with the default environment variables.
 ///
 /// The builder can be configured before being initialized.
+/// This is a convenient way of calling [`Builder::from_default_env`].
+///
+/// [`Builder::from_default_env`]: struct.Builder.html#method.from_default_env
 pub fn builder() -> Builder {
     Builder::from_default_env()
 }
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -1185,13 +1185,20 @@ where
...
+/// [`Builder::from_default_env`]: struct.Builder.html#method.from_default_env
 pub fn builder() -> Builder {
     Builder::from_default_env()
 }
 
 /// Create a builder from the given environment variables.
 ///
 /// The builder can be configured before being initialized.
+#[deprecated(
+    since = "0.7.2",
```
SirWindfield, line 1199: Minor version bump if I am correct. The method has only been deprecated instead of removed, keeping the change backwards compatible.

KodrAus: Ah, we've been following Rust's quasi-standard _minor changes pre-`1.0` are patches_ policy, so as a backwards compatible, non-behavior-breaking change we'd bump this as `0.7.2` instead of `0.8.0`.

KodrAus:
I'll fix up that lint while I'm here.

src/lib.rs
SirWindfield: I meant patch. I’m just too tired haha :)

SirWindfield:
> I'll fix up that lint while I'm here.

Started on #169. Some more popped up after that. You should be able to push to that Pr since you’re maintainer.

SirWindfield:
Anything you want to add to the PR @KodrAus? Else this can be merged.

PR #169, opened 2020-07-19
Fix lints

SirWindfield:


jplatte:
I don't see how moving the `#[allow(missing_docs)]` improves anything. I like the before version more. I don't think anybody would ever have a reason to remove the main type docs. With our MSRV now at 1.41, why are we `#[allow]`ing `clippy::manual_non_exhaustive`?

SirWindfield:
Totally forgot about the Msrv.

And ye you're right. I'll clean up tmrw

SirWindfield:
@jplatte Should we add a type declaration, something like `FormatFn`? It would make most sense to make it public so people can use it as well at this point.

jplatte:
Yeah, if we create a type definition for that (which seems like a good idea) it has to be public, if not by Rust's visibility rules then for people to be able to see its definition in the docs.

jplatte:
PR is a bit outdated by now, squashing and rebasing...

PR #173, opened 2020-07-23
ci(docs): remove ci crate

SirWindfield:


jplatte:
Why did you remove the CI crate from running on CI, but leave it in the repository?

PR #174, opened 2020-07-24
chore(examples): add syslog friendly format

### LINKED ISSUE #157: Systemd format support
I have a Linux service which I run using Systemd. I really like logging to stdout as it is convenient during development.  
Since Systemd supports capturing logs directly from stderr/stdout, I've decided to keep using env_logger rather than the [systemd](https://github.com/jmesmon/rust-systemd) or [syslog](https://github.com/Geal/rust-syslog) crates.

By default the log format is not usable since it creates duplicate information and doesn't keep the log urgency.
```
Mar 16 15:08:01 raspberrypi my_daemon[7697]: [2020-03-16T14:08:01Z WARN  my_daemon::storage] <actual log message...>
```

A better solution is to strip the timestamp and pass the log level to systemd (it uses a `<level>` prefix format):

```rust
fn main() {
    match std::env::var("RUST_LOG_STYLE") {
        Ok(s) if s == "SYSTEMD" => env_logger::builder()
            .format(|buf, record| {
                writeln!(
                    buf,
                    "<{}>{}: {}",
                    match record.level() {
                        log::Level::Error => 3,
                        log::Level::Warn => 4,
                        log::Level::Info => 6,
                        log::Level::Debug => 7,
                        log::Level::Trace => 7,
                    },
                    record.target(),
                    record.args()
                )
            })
            .init(),
        _ => env_logger::init(),
    };
}
```

My feature request is to integrate it natively in env_logger, since I see some synergy here between the two projects.

Feel free to close the issue if you don't think this is a good idea. I'll leave the code snippet here for posterity, maybe someone else will find it useful.

Thanks for env_logger!

SirWindfield:
Closes #157

### DIFF
--- a/examples/syslog_friendly_format.rs
+++ b/examples/syslog_friendly_format.rs
@@ -1,0 +1,24 @@
+use std::io::Write;
+
+fn main() {
+    match std::env::var("RUST_LOG_STYLE") {
+        Ok(s) if s == "SYSTEMD" => env_logger::builder()
+            .format(|buf, record| {
+                writeln!(
+                    buf,
+                    "<{}>{}: {}",
+                    match record.level() {
+                        log::Level::Error => 3,
+                        log::Level::Warn => 4,
+                        log::Level::Info => 6,
+                        log::Level::Debug => 7,
+                        log::Level::Trace => 7,
+                    },
+                    record.target(),
+                    record.args()
+                )
+            })
+            .init(),
+        _ => env_logger::init(),
+    };
+}


SirWindfield:
Lint fail expected.

examples/syslog_friendly_format.rs
```diff
@@ -0,0 +1,24 @@
+use std::io::Write;
+
+fn main() {
+    match std::env::var("RUST_LOG_STYLE") {
+        Ok(s) if s == "SYSTEMD" => env_logger::builder()
```
jplatte, line 5:
```suggestion
    match std::env::var_os("RUST_LOG_STYLE").as_deref() {
        Some(b"SYSTEMD") => env_logger::builder()
```

because `env::var` is lame :smile:

SirWindfield: Feels more natural, totally true. Would you mind changing it in a PR? :)

SirWindfield: I did just merge it as it seemed trivial. Should have waited :)

PR #177, opened 2020-08-14
Support kebab-case names in parse sepc

Gnnng:
This pull request improves the user experience when package names contain hyphens. Users can use either the original form (`my-app`) or the canonical form (`my_app`).

See more about name using hyphens in this [RFC](https://github.com/rust-lang/rfcs/pull/940).


jplatte:
Hey, thanks for the PR! Sorry for the delay, but I personally just don't feel like I should be deciding whether to include this kind of functionality, only being involved in this project for a rather short time.

That said, personally I like this change. The only improvement I can suggest is mentioning this in `README.md` too.

@SirWindfield @KodrAus wdyt?

SirWindfield:
I do like the change! Is it possible that two crates exist with the same name but one uses hyphens, the other dashes? I can only imagine that this can't be the case, as their import names would be the same. This would be the only concern I'd have though.

src/filter/mod.rs
```diff
@@ -613,4 +613,16 @@ mod tests {
         assert_eq!(dirs[0].level, LevelFilter::max());
         assert!(filter.is_some() && filter.unwrap().to_string() == "a*c");
     }
+
+    #[test]
+    fn parse_spec_name_canonicalization() {
+        // accept binary names in both styles (`snake_case` and `kebab-case`)
+        let (dirs, _) = parse_spec("snake_case_crate=info,kebaba-case-crate=debug");
```
jplatte, line 620:
```suggestion
        let (dirs, _) = parse_spec("snake_case_crate=info,kebab-case-crate=debug");
```

jplatte:
@Gnnng I thought it was obvious so didn't leave another review comment, but the typo was replicated in the assertion a few lines below, you need to update that one too :)

Gnnng:
My bad, just fixed it and rebased it. Thanks for reviewing it. Any suggestions on what to put in the `README.md`? It's better if it comes from the maintainers so feel free to amend the PR.

jplatte:
> Any suggestions on what to put in the `README.md`?

I was thinking something similar to what you wrote for `lib.rs`.

KodrAus:
This looks good to me! There shouldn't be any conflicts for names. Cargo accepts either `-` or `_` but considers those characters equal for  determining the name.

jplatte:
@Gnnng Can you rebase again and add this to the readme? If not I might do it, but no promises.

Gnnng:
Sure, let me do that.

SirWindfield:
I can take a closer look an Tuesday, up until then I don't have enough time for a full review :)

SirWindfield (approved):
Looks good. Tests are there and proper documentation has been added to both the readme and the doc comments.

jplatte (approved):
So with the crate name also being overridable replacing `-` by `_` is not always the right conversion from package name to crate name (example: [rust-s3](https://docs.rs/rust-s3/0.26.4/s3/) where the package is named `rust-s3` but the library crate is named `s3`).

I guess this still doesn't break anything though, it could just be expected by people that it does more than it actually does 🤷🏼

SirWindfield:
> 
> 
> So with the crate name also being overridable replacing `-` by `_` is not always the right conversion from package name to crate name (example: [rust-s3](https://docs.rs/rust-s3/0.26.4/s3/) where the package is named `rust-s3` but the library crate is named `s3`, not `rust_s3`).
> 
> I guess this still doesn't break anything though, it could just be expected by people that it does more than it actually does 🤷🏼

We could add a notice to the README with this example to bring attention to these cases :thinking:

PR #192, opened 2020-12-30
Add suffix customization to default formatter.

jthacker:
This defaults to the same suffix as before, \n.
Supports several use cases:
* When writing to a raw TTY, \r\n is required
* Custom formatting of messages may be desirable, e.g. \n\n


KodrAus (approved):
Hi @jthacker! :wave:

Thanks for the PR. This looks good to me!

Issue #178, opened 2020-08-21
Logging to a file?


osa1:
Would it make sense to add a [Target][1] variant for logging to a file? It doesn't make sense to print to stdout or stderr in TUI applications as that breaks the TUI, and having to ask users to redirect stderr to a file is inconvenient.

Context: https://github.com/osa1/tiny/issues/238

[1]: https://docs.rs/env_logger/0.7.1/env_logger/enum.Target.html

jplatte:
Sure, sounds like a simple enough addition. @SirWindfield wdyt?

One thing to note is that this can't be done in a backwards-compatible manner, because `Target` currently implements `Copy`.

jplatte:
If we decide to do this, we might want to copy what [simplelog](https://github.com/drakulix/simplelog.rs) does and accept anything that is writable.

osa1:
> One thing to note is that this can't be done in a backwards-compatible manner, because Target currently implements Copy.

`Send` and `Sync` will also be lost if we simply put a file handle in a `Target` variant, but they can be maintained by wrapping the `File` with `Arc<Mutex<..>>`.

Also, `Target` is currently exposed to users so adding a new variant to it is a breaking change that requires major version bump. So regardless of the traits I think this will require a version bump.

jplatte:
> Also, `Target` is currently exposed to users so adding a new variant to it is a breaking change that requires major version bump. So regardless of the traits I think this will require a version bump.

Right. Could potentially be made non-exhaustive with the next version.

KodrAus:
I’ve previously punted on supporting file logging, because it’s a bit of a can of worms. You can’t really just pipe records into a file, you end up needing to have custom file patterns, rolling files with configurable sizes, deal with transient IO errors and slow writes. It’s definitely possible, and has been done plenty before, but it always ends up being a whole stack of its own to maintain 🙂

jplatte:
@KodrAus What about adding a note that we don't do any of this? I think just piping records into a file would work well for short-lived programs, even if it quickly breaks down for daemons or server applications.

KodrAus:
Maybe we could design it to indirectly support files, instead of baking it in? So it might look something like:

```rust
enum Target {
    Pipe(Box<dyn Write>),
    ..
}
```

which we then wrap in a `Mutex` in the logger and stream records to. You could use a `File` or rolling file wrapper and `env_logger` wouldn’t have to build its own rolling file API?

jplatte:
[...]

PR #195, opened 2021-02-25
Add Target::Pipe

TilCreator:
Now logs can be written to custom pipes and not just to stdout or stderr.

This is my take on #178.

I would be very happy to hear some feedback, since I'm quite unsure about a few changes:
- The [example](examples/custom_target.rs) seams overly complicated, mostly because I didn't want an example to write to a file. Maybe there is a better candidate than `std::sync::mpsc` used as a pipe for that?
- [`Builder::target_pipe`](src/fmt/writer/mod.rs#R110) is my idea for not breaking too much, but maybe there is a solution to implement that into [`Builder::target`](src/fmt/writer/mod.rs#R97) directly?
- In [`Buffer`](src/fmt/writer/termcolor/extern_impl.rs#R100) it is assumed that if `target_pipe` is `Some`, `test_target` is always `None`. Is there a better way to enforce that?

### DIFF
--- a/examples/custom_target.rs
+++ b/examples/custom_target.rs
@@ -18,7 +18,7 @@ $ export MY_LOG_STYLE=never
 #[macro_use]
 extern crate log;
 
-use env_logger::{Builder, Env};
+use env_logger::{Builder, Env, Target};
 use std::{
     io,
     sync::mpsc::{channel, Sender},
@@ -58,7 +58,7 @@ fn main() {
     Builder::from_env(env)
         // The Sender of the channel is given to the logger
         // The wrapper is used, because Sender itself doesn't implement io::Write
-        .target_pipe(WriteAdapter { sender: rx })
+        .target(Target::Pipe(Box::new(WriteAdapter { sender: rx })))
         .init();
 
     trace!("some trace log");
--- a/examples/custom_target.rs
+++ b/examples/custom_target.rs
@@ -1,0 +1,81 @@
+/*!
+Using `env_logger`.
+
+Before running this example, try setting the `MY_LOG_LEVEL` environment variable to `info`:
+
+```no_run,shell
+$ export MY_LOG_LEVEL='info'
+```
+
+Also try setting the `MY_LOG_STYLE` environment variable to `never` to disable colors
+or `auto` to enable them:
+
+```no_run,shell
+$ export MY_LOG_STYLE=never
+```
+*/
+
+#[macro_use]
+extern crate log;
+
+use env_logger::{Builder, Env};
+use std::{
+    io,
+    sync::mpsc::{channel, Sender},
+};
+
+// This struct is used as an adaptor, it implements io::Write and forwards the buffer to a mpsc::Sender
+struct WriteAdapter {
+    sender: Sender<u8>,
+}
+
+impl io::Write for WriteAdapter {
+    // On write we forward each u8 of the buffer to the sender and return the length of the buffer
+    fn write(&mut self, buf: &[u8]) -> io::Result<usize> {
+        for chr in buf {
+            self.sender.send(*chr).unwrap();
+        }
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> io::Result<()> {
+        Ok(())
+    }
+}
+
+fn main() {
+    // The `Env` lets us tweak what the environment
+    // variables to read are and what the default
+    // value is if they're missing
+    let env = Env::default()
+        .filter_or("MY_LOG_LEVEL", "trace")
+        // Normaly using a pipe as target would asume this as false, but this forces it to true
+        .write_style_or("MY_LOG_STYLE", "always");
+
+    // Create the channel for the log messages
+    let (rx, tx) = channel();
+
+    Builder::from_env(env)
+        // The Sender of the channel is given to the logger
+        // The wrapper is used, because Sender itself doesn't implement io::Write
+        .target_pipe(WriteAdapter { sender: rx })
+        .init();
+
+    trace!("some trace log");
+    debug!("some debug log");
+    info!("some information log");
+    warn!("some warning log");
+    error!("some error log");
+
+    // Collect all messages send to the channel and parse the result as a string
+    String::from_utf8(tx.try_iter().collect::<Vec<u8>>())
+        .unwrap()
+        // Split the result into lines so a prefix can be added to each line
+        .split("\n")
+        .for_each(|msg| {
+            // Print the message with a prefix if it has any content
+            if msg.len() > 0 {
+                println!("from pipe: {}", msg)
+            }
+        });
+}
--- a/examples/custom_target.rs
+++ b/examples/custom_target.rs
@@ -49,7 +49,7 @@ fn main() {
     // value is if they're missing
     let env = Env::default()
         .filter_or("MY_LOG_LEVEL", "trace")
-        // Normaly using a pipe as target would asume this as false, but this forces it to true
+        // Normally using a pipe as a target would mean a value of false, but this forces it to be true.
         .write_style_or("MY_LOG_STYLE", "always");
 
     // Create the channel for the log messages
--- a/examples/custom_target.rs
+++ b/examples/custom_target.rs
@@ -57,7 +57,7 @@ fn main() {
 
     Builder::from_env(env)
         // The Sender of the channel is given to the logger
-        // The wrapper is used, because Sender itself doesn't implement io::Write
+        // A wrapper is needed, because the `Sender` itself doesn't implement `std::io::Write`.
         .target(Target::Pipe(Box::new(WriteAdapter { sender: rx })))
         .init();
 
--- a/src/fmt/writer/mod.rs
+++ b/src/fmt/writer/mod.rs
@@ -15,10 +15,41 @@ pub(in crate::fmt) mod glob {
 
 pub(in crate::fmt) use self::termcolor::Buffer;
 
-/// Log target, either `stdout` or `stderr`.
-#[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+/// Log target, either `stdout`, `stderr` or a custom pipe.
 #[non_exhaustive]
 pub enum Target {
+    /// Logs will be sent to standard output.
+    Stdout,
+    /// Logs will be sent to standard error.
+    Stderr,
+    /// Logs will be sent to a custom pipe.
+    Pipe(Box<dyn io::Write + Send + 'static>),
+}
+
+impl Default for Target {
+    fn default() -> Self {
+        Target::Stderr
+    }
+}
+
+impl fmt::Debug for Target {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::result::Result<(), std::fmt::Error> {
+        write!(
+            f,
+            "{}",
+            match self {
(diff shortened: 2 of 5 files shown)


TilCreator:
I can't reproduce the failed check locally and I also didn't change anything on the code that is failing. Plz help?

jplatte:
It's `cargo clippy`, not `cargo check`, that's failing (and with a nightly version of clippy). I've pushed a commit to master that fixes that lint plus one more, rebasing should get rid of the CI failure now.

TilCreator:
Ah, I wasn't using the nightly version of `clippy`. I rebased master and the checks succeed now. thx

src/fmt/writer/mod.rs
```diff
@@ -14,11 +17,14 @@ pub(in crate::fmt) use self::termcolor::Buffer;
 
 /// Log target, either `stdout` or `stderr`.
 #[derive(Clone, Copy, Debug, Eq, Hash, PartialEq)]
+#[non_exhaustive]
 pub enum Target {
     /// Logs will be sent to standard output.
     Stdout,
     /// Logs will be sent to standard error.
     Stderr,
+    /// Logs will be send to a custom pipe.
```
jplatte, line 26:
```suggestion
    /// Logs will be sent to a custom pipe.
```

jplatte:
You did the rebase the wrong way around (you should rebase your branch on top of master, not the other way around).

> [`Builder::target_pipe`](src/fmt/writer/mod.rs#R110) is my idea for not breaking too much, but maybe there is a solution to implement that into [`Builder::target`](src/fmt/writer/mod.rs#R97) directly?

What do you mean by "not breaking too much"? The derived traits for `Target`? I think I'd rather see those removed than there being a separate `target_pipe` builder method / field. I wonder why e.g. `Hash` is implemented for that in the first place.

jplatte:
By the way, thank you for working on this!

TilCreator:
Ups, I knew there was something wrong after at least the 5th git command, is it still worth trying to fix? ^^

jplatte:
It should be as easy as

```sh
// if you haven't set up the remote yet, though I guess you have
git remote add upstream https://github.com/env-logger-rs/env_logger
git fetch upstream
// rebase your branch on top of upstream's master branch
git rebase upstream/master
```

even now. (git should detect the duplicate commit on your branch and get rid of it)

Followed by `git push --force` to tell git you really want to rewrite the history of the remote branch.

TilCreator:
I have tried myself on keeping the pipe in the `Target::Pipe`, but [that](https://github.com/TilCreator/env_logger/commit/9e2f55a4e8be24fa88628f7b56d716151ee3f67e) is just such a mess, so many `Arc<Mutex<...>>`....
`Target` also gets used a lot after `build()` to check if stdout or stdout is used, so removing or "circumenting" the derived traits of `Target` is hard.

I will try something that takes the value out of `Target::Pipe` on `target()` or `build()` tomorrow.
Maybe I can even come up something better than `enum Target {..., Pipe(Option<Box<dyn io::Write + Send + 'static>>)}` after getting some sleep ^^

TilCreator:
I'm quite happy with this version as it breaks very little if anything at all and it feels like a simple extension of the old `target` method.
Now I just need a better example as the [current one](examples/custom_target.rs) is just confusing and not super practical.

SirWindfield:
I can take a closer look an Tuesday, up until then I don't have enough time for a full review :)

src/fmt/writer/termcolor/shim_impl.rs
```diff
@@ -28,17 +44,27 @@ impl BufferWriter {
...
 
-        Ok(())
+            match self.target {
+                TargetType::Stderr => eprint!("{}", log),
+                TargetType::Stdout => print!("{}", log),
+                TargetType::Pipe => unreachable!(),
+            }
+
+            Ok(())
+        }
```
SirWindfield, line 67: You could merge the two (logic) branches into a single match arm:

````suggestion
pub(in crate::fmt::writer) fn print(&self, buf: &Buffer) -> io::Result<()> {
    // This impl uses the `eprint` and `print` macros
    // This is so their output can be captured by `cargo test`.
    match self.target {
        // Safety: If the target type is `Pipe`, `target_pipe` will always be non-empty.
        TargetType::Pipe => self
            .target_pipe
            .as_ref()
            .unwrap()
            .lock()
            .unwrap()
            .write_all(&buf.0)?,
        TargetType::Stdout => print!("{}", String::from_utf8_lossy(&buf.0)),
        TargetType::Stderr => eprint!("{}", String::from_utf8_lossy(&buf.0)),
    }

    Ok(())
}
````

As far as I can tell the safety does apply to the PR.

File: examples/custom_target.rs [hunk 3]
```diff
@@ -0,0 +1,81 @@
...
+Also try setting the `MY_LOG_STYLE` environment variable to `never` to disable colors
+or `auto` to enable them:
+
+```no_run,shell
+$ export MY_LOG_STYLE=never
+```
+*/
+
+#[macro_use]
+extern crate log;
```
SirWindfield, line 19: 2018 idoms, you should be able to just call `use log::{info, error, warn};`. Is there a policy in-place for examples and what edition they should target? If not, I would use the newer syntax.

examples/custom_target.rs
```diff
@@ -0,0 +1,81 @@
...
+    }
+}
+
+fn main() {
+    // The `Env` lets us tweak what the environment
+    // variables to read are and what the default
+    // value is if they're missing
+    let env = Env::default()
+        .filter_or("MY_LOG_LEVEL", "trace")
+        // Normaly using a pipe as target would asume this as false, but this forces it to true
```
SirWindfield, line 52:
```suggestion
        // Normally using a pipe as a target would mean a value of false, but this forces it to be true.
```

examples/custom_target.rs
```diff
@@ -0,0 +1,81 @@
...
+        .write_style_or("MY_LOG_STYLE", "always");
+
+    // Create the channel for the log messages
+    let (rx, tx) = channel();
+
+    Builder::from_env(env)
+        // The Sender of the channel is given to the logger
+        // The wrapper is used, because Sender itself doesn't implement io::Write
```
SirWindfield, line 60:
```suggestion
        // A wrapper is needed, because the `Sender` itself doesn't implement `std::io::Write`.
```

src/lib.rs
```diff
@@ -708,7 +708,10 @@ impl Builder {
 
     /// Sets the target for the log output.
     ///
-    /// Env logger can log to either stdout or stderr. The default is stderr.
+    /// Env logger can log to either stdout, stderr or a custom pipe. The default is stderr.
+    ///
+    /// The custom pipe can be used to send the log to a file directly or something more complex.
+    /// It is advertised to use a wrapper for log files, so that rollover and slow FSs are handled well.
```
SirWindfield, line 714:
```suggestion
    /// The custom pipe can be used to send the log messages to a custom sink (for example a file).
    /// Do note that direct writes to a file can become a bottleneck due to IO operation times.
```

I don't think a notice is needed to be honest, but if we add one I'd go for something similar to the above without going into implementation details.

SirWindfield (changes requested):
The rest looks decent. Maybe @jplatte could look over again as he was the initial reviewer, and a second pair of eyes is always good ;)

That being said, sorry that it took me that long.

src/fmt/writer/termcolor/shim_impl.rs
TilCreator: Ah fluff, I applied the suggested changes backwards and now I can't apply this one directly. (I'm a bit new to reviews)

TilCreator:
thx @SirWindfield (also for fixing my typos ^^)

examples/custom_target.rs
TilCreator: I would agree that this should be changed, but all other examples are very similar and use this line. So a bulk change of all the examples would be better. Should I still change it?

jplatte:
I can have a second look, sure. Probably going to take a while for me as well.

jose-acevedoflores:
I've been playing around with this pr since I need support for logging to a file and this works great for my use case!

I did have one question about the possibility of having multiple targets like stdout and a file. Is this something out of scope for the goals of this library or is it something that could be entertained ?

TilCreator:
(I'm just a contributor, not a maintainer)
I think this is possibly better implemented by the user, it sounds like a quite specific use case. (Maybe this could be added to the [example](examples/custom_target.rs) (and I'm still not entirely happy with the example)?)

src/fmt/writer/mod.rs
```diff
@@ -68,7 +115,8 @@ impl Writer {
 ///
 /// The target and style choice can be configured before building.
 pub(crate) struct Builder {
-    target: Target,
+    target_type: TargetType,
+    target_pipe: Option<Arc<Mutex<dyn io::Write + Send + 'static>>>,
```
jplatte, line 119: Why split the target into two fields and introduce a new `TargetType` type? I think `target` can just be wrapped in `Option` for the `build` method to work (where you can then use `self.target.take().unwrap_or_default()` to get the actual value).

src/fmt/writer/termcolor/extern_impl.rs
```diff
@@ -70,46 +71,72 @@ impl Formatter {
 
 pub(in crate::fmt::writer) struct BufferWriter {
     inner: termcolor::BufferWriter,
-    test_target: Option<Target>,
+    test_target_type: Option<TargetType>,
+    target_pipe: Option<Arc<Mutex<dyn io::Write + Send + 'static>>>,
```
jplatte, line 75: Can you explain why `Arc<Mutex<>>` is needed?

src/fmt/writer/termcolor/shim_impl.rs
```diff
@@ -30,14 +46,19 @@ impl BufferWriter {
...
+                .target_pipe
+                .as_ref()
+                .unwrap()
+                .lock()
+                .unwrap()
+                .write_all(&buf.0)?,
+            TargetType::Stdout => print!("{}", String::from_utf8_lossy(&buf.0)),
+            TargetType::Stderr => eprint!("{}", String::from_utf8_lossy(&buf.0)),
         }
-
```
jplatte, line 40: Maybe put back the empty line before the final expression of the function.

src/fmt/writer/mod.rs
TilCreator: I did try that, because I would like it way more that way, too. But the Target gets passed (and cloned) everywhere (for color checking and the `shim_impl`) that is why `TargetType` needs `Clone, Copy, Debug, Eq, Hash, PartialEq` implemented which is hard to impossible with `dyn io::Write + Send + 'static`.

src/fmt/writer/termcolor/extern_impl.rs
TilCreator: Because the pipe is still partly owned by the `Builder` (it doesn't get deconstructed on `build()`).

src/fmt/writer/termcolor/shim_impl.rs
TilCreator: That one just got lost ^^

jplatte:
I'm still planning to have a look at this again, hopefully this or next week.

jplatte:
Merged in #204.

Issue #196, opened 2021-03-31
Builder#parse_env does not override manually set filter level


BramBonne:
Consider the following example ([runnable version](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=db7652ef4104ae10fbdadc97b4c74ce6)):
```rust
std::env::set_var("RUST_LOG", "debug");

let mut builder = env_logger::Builder::new();
builder.filter_level(log::LevelFilter::Trace);

// The following call should override the log level to Debug, as read from the environment
builder.parse_default_env();
    
builder.init();
assert_eq!(log::max_level(), log::LevelFilter::Trace);
```

According to the [documentation](https://docs.rs/env_logger/0.8.3/env_logger/struct.Builder.html#method.parse_env), the `parse_env` function "allows a builder to be configured with default parameters, to be then overridden by the environment". For this example, that would mean that `log::max_level()` should be set to `Debug` at the end of this example. In reality, it's set to `Trace` instead.

Removing the call to `builder.filter_level(log::LevelFilter::Trace)` correctly sets the log level to `Debug` instead.

jplatte:
Yeah, the docs are pretty clear here. A PR fixing this would be welcome.

BramBonne:
I provided both #199 and #200 as alternative potential fixes to this problem, depending on whether we want directive names to be unique in general (#200), or not (#199).

BramBonne:
Hey @jplatte, friendly ping on this. Would #200 be an acceptable way to fix this, or is there a better approach? Thanks for taking a look :-)

Kind regards,
Bram

SirWindfield:
I'll have time on the weekend to take a look at the PR.

BramBonne:
Thank you! But no rush, I was just checking it was still being considered 🙂

SirWindfield:
@jplatte I've taken a look at both of the PRs. Both look good.

I don't have strong preference for either PR, but if I'd have to choose I'd say we should merge #200, as it means more cosistent behaviour across the library in general by __always__ overwriting previous directives. I can't think of an example off the top of my head where this might not be wanted.

jplatte:
I've had a look, I agree that #200 seems like the better solution.

BramBonne:
Hey,

Just checking if anything is expected/needed from my end here.

Kind regards,
Bram

jplatte:
No. I merged #200. Sorry about the long delay.

PR #199, opened 2021-04-08
Override existing directives when parsing

BramBonne:
This should work as a minimal fix for #196.

Let me know if you would like `filter::Builder::filter()` to behave in the same way (i.e., removing existing directives with the same name). Also, please let me know if this fix doesn't make sense -- I'm not familiar with this code, and even just starting out with Rust as well.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -220,6 +220,10 @@ impl Builder {
 
         self.filter = filter;
 
+        // Overwrite existing directives with the same name.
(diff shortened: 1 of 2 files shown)


SirWindfield (approved):

SirWindfield:
DON'T merge until it is decided which PR should get into master (#199 or #200).

jplatte:
#200 was merged.

PR #200, opened 2021-04-09
Ensure unique directive names when building filters

BramBonne:
This is an alternative fix for #196, if we are OK with all later calls to the various `filter::Builder` methods taking precedence over earlier calls.


SirWindfield (approved):

SirWindfield:
DON'T merge until it is decided which PR should get into master (#199 or #200).

PR #204, opened 2021-05-01
Add Target::Pipe cleaned up

jplatte:
#195 with some adjustments to have the pipe as enum variant data instead of a separate field.

### DIFF
--- a/src/fmt/writer/termcolor/extern_impl.rs
+++ b/src/fmt/writer/termcolor/extern_impl.rs
@@ -3,6 +3,7 @@ use std::cell::RefCell;
 use std::fmt;
 use std::io::{self, Write};
 use std::rc::Rc;
+use std::sync::{Arc, Mutex};
 
 use log::Level;
 use termcolor::{self, ColorChoice, ColorSpec, WriteColor};
@@ -71,6 +72,7 @@ impl Formatter {
 pub(in crate::fmt::writer) struct BufferWriter {
     inner: termcolor::BufferWriter,
     test_target: Option<Target>,
+    target_pipe: Option<Arc<Mutex<dyn io::Write + Send + 'static>>>,
 }
 
 pub(in crate::fmt) struct Buffer {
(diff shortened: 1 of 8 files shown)


src/fmt/writer/termcolor/extern_impl.rs
```diff
@@ -70,46 +71,71 @@ impl Formatter {
...
 pub(in crate::fmt::writer) struct BufferWriter {
     inner: termcolor::BufferWriter,
-    test_target: Option<Target>,
+    test_target: Option<WritableTarget>,
 }
 
 pub(in crate::fmt) struct Buffer {
     inner: termcolor::Buffer,
-    test_target: Option<Target>,
+    test_target: bool,
```
jplatte, line 79: This change I'm not so sure about. I haven't really tried to understand the surrounding code, but it was the easiest fix.

I am confident that the behavior is still the same as before, but maybe there was some reason the test target was being stored before? Also, having the same field name as in `BufferWriter` but a different type seems like a code smell, but I couldn't really come up with something better due to lack of understanding of the code overall.

jplatte:
@TilCreator please have a look.

cc @SirWindfield, maybe you want to review the overall changes again?

src/fmt/writer/termcolor/extern_impl.rs
TilCreator: At first I was a bit confused, but I can not find anything wrong and I think it's an improvement.
Maybe it would be less confusing to rename `Buffer::test_target` to `Buffer::is_test`?

TilCreator (approved):

src/fmt/writer/termcolor/extern_impl.rs
jplatte: I'll check again how it's used and rename either to `is_test` or sth. like `has_test_target`.

SirWindfield:
Sure, I'll do it on Saturday! Sorry for the long delay, university has been eating up all my time.

SirWindfield:
Doing it this evening, sorry for the delay :(

src/fmt/writer/termcolor/extern_impl.rs
SirWindfield: Ye, I do agree on the `has_test_target`.

SirWindfield (approved):
Looks good to me! 
I am still confused on why `clippy` didn't pick up the `len() > 0` statements and recommended `is_empty()` instead. Pretty sure there was a lint for that.

Issue #179, opened 2020-09-30
Log with target instead of module path by default?


inferiorhumanorgans:
The`log` macro sets the target to the module path by default but only allows users to override the target.  By using the target instead of the module path in the default formatter an end user could trivially modify the default format.

Thoughts?

jplatte:
I'm not certain I understand the full implications of this change. @KodrAus wdyt?

KodrAus:
It sounds like it is confusing that we're using the module path instead of the target, especially since they're _usually_ the same.

I'd be happy to change the default format to use the target instead.

PR #205, opened 2021-05-06
use target instead of module_path in formatter by default

### LINKED ISSUE #155: Module path and target confusion
The “Enabling logging” section of the documentation states that the module path is used to filter logging statements. But in fact the target of a logging record  (which defaults to the module path) is used for filtering.

In addition the default formatter prints the module path and not the target. This lead to some confusion for me when I tried to write a filter based on the output I was seeing.

It would be great if the documentation would be corrected and ideal if the target was printed by the default formatter.

xofyarg:
Target is used in filter, and its value defaults to module_path. To
enable user customize the format, also to reduce confusion, makes
formatter supporting target as well.

This change introduces a boolean builder variable "honor_target". When
set to true(the default), formatter will use target in the record
metadata.

Fix #155 and #179.

### DIFF
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -141,6 +141,7 @@ pub(crate) type FormatFn = Box<dyn Fn(&mut Formatter, &Record) -> io::Result<()>
 pub(crate) struct Builder {
     pub format_timestamp: Option<TimestampPrecision>,
     pub format_module_path: bool,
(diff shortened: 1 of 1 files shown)


deastoe:
I'd love to see this feature, even if the default behaviour remains the same.

mainrs:
Superseeded by #209.

PR #206, opened 2021-06-11
Bump version to 0.8.4

BramBonne:
Suggestion to bump the version, as merging #200 will cause slight behavior changes to users.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,7 @@
 [package]
 name = "env_logger"
 edition = "2018"
-version = "0.8.3"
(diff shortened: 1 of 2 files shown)


jplatte:
Thanks, published.

jplatte:
Actually this was technically a breaking change release, since we introduced `Target::Pipe` (and `Target` didn't previously have the `#[non_exhaustive]` attribute) and removed a bunch of `Target`s derived trait implementations. I'm not sure it's worth to yank and release as 0.9.0 unless somebody reports issues with the release though.

PR #209, opened 2021-07-03
Show target instead of module path by default

### LINKED ISSUE #155: Module path and target confusion

### LINKED ISSUE #179: Log with target instead of module path by default?

gtsiam:
This is an alternative to #205, since that pr seems to be incomplete/stalled.

- d2998a6 Added the option of printing the log target in the header
- 365ffaf Changed defaults:
  - Module path is now hidden by default
  - Target is now shown by default
- 1888497 Clarified the difference between module path and log target

Fixes #155
Fixes #179

These changes also likely warrant a minor version bump since they alter the default output format - but I'm leaving it out of the pr since I'm unsure of what the exact roadmap is, if any.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -238,7 +238,7 @@ impl Builder {
             });
         } else {
             // Consume map of directives.
-            let directives_map = mem::replace(&mut self.directives, HashMap::new());
+            let directives_map = mem::take(&mut self.directives);
             directives = directives_map
                 .into_iter()
                 .map(|(name, level)| Directive { name, level })
@@ -253,7 +253,7 @@ impl Builder {
         }
 
         Filter {
-            directives: mem::replace(&mut directives, Vec::new()),
+            directives: mem::take(&mut directives),
             filter: mem::replace(&mut self.filter, None),
         }
     }
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -153,8 +153,8 @@ impl Default for Builder {
     fn default() -> Self {
         Builder {
             format_timestamp: Some(Default::default()),
-            format_module_path: true,
-            format_target: false,
+            format_module_path: false,
+            format_target: true,
             format_level: true,
             format_indent: Some(4),
             custom_format: None,
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -141,6 +141,7 @@ pub(crate) type FormatFn = Box<dyn Fn(&mut Formatter, &Record) -> io::Result<()>
 pub(crate) struct Builder {
     pub format_timestamp: Option<TimestampPrecision>,
     pub format_module_path: bool,
+    pub format_target: bool,
     pub format_level: bool,
     pub format_indent: Option<usize>,
     pub custom_format: Option<FormatFn>,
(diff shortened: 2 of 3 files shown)


jplatte:
> that pr seems to be incomplete/stalled

I think the main problem is there is nobody *actively* maintaining this crate. I'm helping with PRs that I consider uncontroversial, but for #205 I have no idea about that / whether it could break some people.

> These changes also likely warrant a minor version bump since they alter the default output format - but I'm leaving it out of the pr since I'm unsure of what the exact roadmap is, if any.

Yeah, the version should be bumped by a maintainer when doing a release, not ahead of time.

gtsiam:
> there is nobody actively maintaining this crate

Yeah, I got that kind of feeling browsing through the commit history... However it's still in heavy use, so I think this is worth the effort.

> I have no idea about that / whether it could break some people

I can remove the second commit and this shouldn't break compatibility in any way, if you prefer. As far as the public api is concerned it would only add the function `Builder::format_target(bool)` which, at least for me, would be very convenient. (well, and the documentation clarification)

I originally included 365ffaf, due to the comments stating willingness to do just that in the linked issues (by @KodrAus), but it would be easy to keep the current default, and opt in to the new behavior until it's convenient/decided to make it the new default:

```rust
env_logger::builder()
    .format_target(true)
    .format_module_path(false)
    .init();
```

jplatte:
You seem to be dedicated to making this work, which is nice. However, I really don't want to dedicate more than an absolute minimum of my time to maintenance here¹. @SirWindfield if you're still around it would be nice if you could decide whether / how to move forward here.

<details><summary>¹</summary>Mainly since I've fully switched to tracing and so am no longer using `env_logger`.</details>

gtsiam:
Context for this is [`warp::log()`](https://docs.rs/warp/0.3.1/warp/filters/log/fn.log.html). I just wanted to see my requests with the target I set, instead of the generic module path.

I'd heard of `tracing`, but it seemed too complex/incompatible for my use-case, so I just went with `env_logger`. But if the maintenance situation is really _that_ bad I'll take the time and look at switching to it (I just found out about `tracing-log` too, so there shouldn't be any problems).

For now I'll just leave this here. If you need me to change anything, ping me. Otherwise, I guess just let it be.

But this is a simple change, all things considered. If maintaining the crate is too time-consuming and there is a good alternative, It may be time to consider mentioning that in the readme and/or the docs...

SirWindfield:
I'm in the same boat. Uni takes most of my time right now (finals 🥳). I have to take a day to see what exatly changes behaviour-vise in this and other PRs. 
I don't want to make promises that I can't hold though. I've blocked my Saturday morning next week after one of my exams. I can take a look but I highly doubt that it would be enough time to fully get a merge going. It's atleast I can give some feedback and thoughts on the feature :)

Sorry for the lack of activity.

mainrs:
Looking good to me

mainrs (approved):

PR #211, opened 2021-07-10
Fix #208 (Remove pipe for BufferWriter and implement handeling in Writer)

### LINKED ISSUE #208: Target::Pipe doesn't work
In #204, we introduce `Target::Pipe` to redirect log messages into custom writer. And it's included in 0.8.4 now.
But when I try [the example in that PR](https://github.com/env-logger-rs/env_logger/pull/204/files#diff-24824a94b419dd85f77d13b3858f55f97857a3093a7a862d180d54a83786ad6b), it doesn't print the message `from pipe` at all, and not even call `WriteAdapter::write`. The `log::error` message just go straight to stderr.
How to make it catch all log message into my custom writer object?

TilCreator:
Fixes #208

A test would also be nice, so that it hopefully doesn't break again unnoticed. ^^

### DIFF
--- a/src/fmt/writer/mod.rs
+++ b/src/fmt/writer/mod.rs
@@ -105,6 +105,7 @@ impl Default for WriteStyle {
 pub(crate) struct Writer {
     inner: BufferWriter,
     write_style: WriteStyle,
+    target_pipe: Option<WritableTarget>,
 }
 
 impl Writer {
@@ -117,7 +118,11 @@ impl Writer {
     }
 
     pub(super) fn print(&self, buf: &Buffer) -> io::Result<()> {
-        self.inner.print(buf)
+        match &self.target_pipe {
+            Some(WritableTarget::Pipe(pipe)) => pipe.lock().unwrap().write_all(&buf.bytes()),
+            Some(_) => unreachable!(),
+            None => self.inner.print(buf)
+        }
     }
 }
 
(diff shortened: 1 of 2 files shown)


mainrs:
I'll write a test today and merge.

TilCreator:
Sry for the force push, but at lest the linter fixes are cleaner now.

src/fmt/writer/termcolor/shim_impl.rs
```diff
@@ -42,10 +32,9 @@ impl BufferWriter {
         // instead of using the streams directly.
         // This is so their output can be captured by `cargo test`.
         match &self.target {
-            // Safety: If the target type is `Pipe`, `target_pipe` will always be non-empty.
-            WritableTarget::Pipe(pipe) => pipe.lock().unwrap().write_all(&buf.0)?,
             WritableTarget::Stdout => print!("{}", String::from_utf8_lossy(&buf.0)),
             WritableTarget::Stderr => eprint!("{}", String::from_utf8_lossy(&buf.0)),
+            WritableTarget::Pipe(_) => unreachable!(),
```
mainrs, line 37: Maybe I am missing something quite obvious, but why is this unreachable?

TilCreator: `BufferWriter::pipe` was removed, so there is only `BufferWriter::stdout` and `BufferWriter::stderr` left to create a `BufferWriter` and both of those can't have the Pipe Target.

jplatte:
Ping @mainrs.

TilCreator:
Ping @mainrs

PR #215, opened 2021-09-06
Add `Style::set_dimmed` to support dimmed text

LukasKalbertodt:
Dimmed text is widely supported and has been added to termcolor in 1.1.1 in this PR: https://github.com/BurntSushi/termcolor/pull/34

Bumping the required `termcolor` version shouldn't be a problem, because it's only a minor bump.


LukasKalbertodt:
Would it be possible to release this as 0.9.1? The last release is 11 months ago and it seems like all commits from `main` could easily be released as a minor new version. That'd be sweet :) No stress though.

LukasKalbertodt:
I totally understand as I struggle with the exact same thing myself for many projects I started. Consequently, I don't have any spare time to help out here, sorry :/

PR #217, opened 2021-10-08
Try to fix Target::Pipe

jplatte:
I thought I found a good way of dealing with this, but it didn't work out at all. I got it to almost compile but now I found that there's another set of things that all rely on there being a termcolor output buffer. (everything around `fmt::Formatter`)

@TilCreator if are you still interested in getting `Target::Pipe` fixed, maybe you can have a look at what it would take to make this work.


jplatte:
Closing in favor of #239.

PR #222, opened 2021-11-11
Fix copy-paste error in docs

Felerius:
The description for the example of `Builder::filter_level` was referencing a module path even though `filter_level` is applied globally.


jplatte (approved):
Thanks!

PR #235, opened 2022-06-22
Update regex version to fix security bug in regex crate

### LINKED ISSUE #230: Regex version out of date, perhaps run a cargo upgrade?
Github audit is reporting a vulnerability with the current version of regex in this package.
Could either enable dependabot, or run a cargo upgrade pass. I would make a PR, but there are a lot of active PRs with no progress.

@KodrAus @mainrs 

Thanks!

azuchi:
Closes #230

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ members = [
 
 [dependencies]
 log = { version = "0.4.8", features = ["std"] }
-regex = { version = "1.0.3", optional = true, default-features=false, features=["std", "perf"] }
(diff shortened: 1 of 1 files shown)


mainrs (approved):
https://blog.rust-lang.org/2022/03/08/cve-2022-24713.html mentions that the problem has been fixed for >1.5.5.

jplatte:
We don't need to bump the minimum versions to fix this for users – see https://github.com/env-logger-rs/env_logger/issues/230#issuecomment-1206265567.

PR #237, opened 2022-09-17
Update listed env_logger and log versions in Readme

atouchet:
Update listed env_logger version to latest release and update log to version required in Cargo.toml.


jplatte:
Doesn't really make a difference for semver-compatible versions, Cargo will automatically use the latest.

PR #239, opened 2022-10-03
Issue 208 pipe fix

niklasha:
Make Pipe use uncolored semantics always.

### DIFF
--- a/src/fmt/writer/termcolor/extern_impl.rs
+++ b/src/fmt/writer/termcolor/extern_impl.rs
@@ -103,7 +103,7 @@ impl BufferWriter {
     }
 
     pub(in crate::fmt::writer) fn pipe(
-        is_test: bool,
+        _is_test: bool,
         write_style: WriteStyle,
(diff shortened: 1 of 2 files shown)


src/fmt/writer/termcolor/extern_impl.rs
```diff
@@ -103,30 +103,26 @@ impl BufferWriter {
     }
 
     pub(in crate::fmt::writer) fn pipe(
-        is_test: bool,
+        _is_test: bool,
```
jplatte, line 106: If this parameter is no longer used, it should be removed.

niklasha: OK done.

jplatte (approved):

jplatte:
Sorry for the long delay on this. Since CI isn't running, I will check everything locally and release then.

jplatte:
Merged as 84b701dac94e4135f59f069a8c155bb9b7922788 and released as v0.9.2.

PR #241, opened 2022-11-05
perf: Replace `HashMap` with a `Vec`

ChrisDenton:
This may be a minor startup perf win (as `env_logger` does not need the expense of DoS resistant hashes) but my main reason for doing this is that the default hashing algorithm depends on random number generation, which may fail. This is important to avoid especially because `env_logger` is often used on program startup. This was a real issue on Windows. While that specific issue is now fixed, it would be nice to avoid this class of issues if possible.

### DIFF
--- a/src/filter/mod.rs
+++ b/src/filter/mod.rs
@@ -59,7 +59,6 @@
 //! [`Filter::matches`]: struct.Filter.html#method.matches
 
 use log::{Level, LevelFilter, Metadata, Record};
-use std::collections::HashMap;
 use std::env;
 use std::fmt;
 use std::mem;
@@ -108,7 +107,7 @@ pub struct Filter {
 ///
 /// [`Filter`]: struct.Filter.html
 pub struct Builder {
-    directives: HashMap<Option<String>, LevelFilter>,
+    directives: Vec<Directive>,
     filter: Option<inner::Filter>,
     built: bool,
 }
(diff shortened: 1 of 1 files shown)


src/filter/mod.rs
```diff
@@ -189,6 +188,16 @@ impl Builder {
         builder
     }
 
+    /// Insert the directive replacing any directive with the same name.
+    fn insert_directive(&mut self, mut directive: Directive) {
+        let by_name = |d: &Directive| d.name == directive.name;
+        if let Some(pos) = self.directives.iter().position(by_name) {
```
matthiasbeyer, line 194: Any particular reason why you didn't write that as

```suggestion
        if let Some(pos) = self.directives.iter().position(|d: &Directive| d.name == directive.name) {
```

?

matthiasbeyer (changes requested):
Hi!

I have a question (read below).

Maybe also: Can you insert the rationale from your PR message in the commit message? This way people finding this to be a `Vec` instead of a `HashMap` don't have to search github to find why it was written this way!

:+1:

src/filter/mod.rs
ChrisDenton: Hi, sorry for the slow reply. Just catching up after holiday + illness. I believe I chose to do this due to the way rustfmt splits up the long `if let` line:

```suggestion
        if let Some(pos) = self
            .directives
            .iter()
            .position(|d: &Directive| d.name == directive.name)
        {
```

Though I'm happy with either suggestion. Either way I'll rebase and add to the commit comment.

epage (approved):

PR #246, opened 2022-11-22
chore: replace atty with is-terminal

jcgruenhage:
`atty` is unmaintained and has a potential unaligned read. See https://github.com/rustsec/advisory-db/blob/main/crates/atty/RUSTSEC-2021-0145.md.

`is-terminal` is a replacement based on `atty`, with the soundness issue fixed and an (IMO) nicer to use API, mirroring what's available in the std lib on nightly with `std::io::IsTerminal`.


epage:
This is a breaking change and is being superseded by #248

PR #255, opened 2022-12-03
 Use cargo add instead of a toml snippet in README

rnestler:


direc85:
Related: I can't install `env_logger = "0.10.0"` because candidate versions include `0.9.3, 0.9.2, 0.9.1, ...` so there seems to be something wrong with publishing the package.

README.md
```diff
@@ -19,7 +19,7 @@ It must be added along with `log` to the project dependencies:
 ```toml
 [dependencies]
 log = "0.4.0"
-env_logger = "0.9.0"
+env_logger = "0.10.0"
 ```
```
epage, line 23: Can you instead change this to
```console
$ cargo add log env_logger
```

And if we do keep the current form, we should also update the release replacements in `Cargo.toml` so this happens automatically on release.

epage:
Sorry, somehow I lost track of this PR

> Related: I can't install `env_logger = "0.10.0"` because candidate versions include `0.9.3, 0.9.2, 0.9.1, ...` so there seems to be something wrong with publishing the package.

What version of Rust are you using?  If its before 1.60, then that is below the MSRV for 0.10 and the reason you aren't getting a better error message is due to https://github.com/rust-lang/cargo/issues/10623

README.md
rnestler: Done.

epage:
Thanks!

Issue #274, opened 2023-08-09
`WriteStyle` is ignored when used with `Target::Pipe`


miraclx:
The patch https://github.com/rust-cli/env_logger/pull/239, while addressing issue https://github.com/rust-cli/env_logger/issues/208 made it so we can write to a custom sink without being in a test context.

The unfortunate downside of that implementation is it disregards the `WriteStyle` specification.

No colors when piping to a custom sink is an acceptable default, but when the user specifies `WriteStyle::Always`, we should have colors written into our sink.

PR #275, opened 2023-08-09
fix: `Target::Pipe` respects `WriteStyle`

miraclx:
Resolves https://github.com/rust-cli/env_logger/issues/274.

The old behavior made it impossible to retain colors in custom sinks.

With this patch, the default behavior assumes non-color compliance with the sink. But this can be overridden with `WriteStyle::Always`.

### DIFF
--- a/src/fmt/writer/mod.rs
+++ b/src/fmt/writer/mod.rs
@@ -199,7 +199,7 @@ impl Builder {
 
         Writer {
             inner: writer,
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
What exactly does this PR change? I only see renamed variables, a cast that I don't think is necessary and some `if {} else {}` changed to `then_some()`?

epage:
This is superseded by #298

Issue #276, opened 2023-08-24
Replace is-terminal dep with std::io::IsTerminal


LunNova:
https://doc.rust-lang.org/stable/std/io/trait.IsTerminal.html

Needs to wait for MSRV 1.70

epage:
Looks like we don't have a state MSRV policy.  My default choice is N-2 but I suspect that env_logger's wider use might make that unappealing. 

@matthiasbeyer any thoughts?

matthiasbeyer:
Yeah so I would actually keep this crate as backwards compatible as possible within reason. If we currently compile on N-10 or even older, I don't see why we should change that if we do not have some really good reasons.

Rationale is that this is, IMO, the go-to crate for "I want logging". It's the lowest-hanging fruit for getting logging to work in a Rust project, at least in my perception.

I would like to see this issue resolved, but I think backwards compatibility is more important.

I'm open to discussion though.

LunNova:
Might be possible to use `cfg(not(version("1.70")))` to use that dep only on old versions and avoid bumping MSRV. Haven't tried, not sure if you can do that in Cargo.toml.

epage:
Another change I'm looking to make, switching from `termcolor` to `anstream`, would require an MSRV bump and would lead to a breaking change (removal of the env_logger styling API).

While I'm not a fan of treating MSRV as a breaking change, if its done in conjunction with breakings changes (with a commitment to support old versions if needed), I wonder if that would be sufficient to allow us to move forward with this.

kayabaNerve:
You can write a build script such that it sets a feature for either is-terminal or std-io-isterminal based on the version in use. That wouldn't be breaking.

epage:
I've gone ahead with my idea of bundling the MSRV bump with a pre-planned breaking change of switching from `termcolor` to `anstream`.  Longer term, I'm hoping rust-lang/rfcs#3537 will help reduce the conflict between the users who want the latest and those who need an earlier version.

See #298

PR #280, opened 2023-10-03
Use `std::io::IsTerminal` rather than `is-terminal`

joshtriplett:
This removes many large dependencies.


epage:
This is being discussed in #276 with the main concern being the MSRV

epage:
This is superseded by #298
(bypassing the MSRV discussion by bundling this in with a breaking change)

PR #285, opened 2023-11-12
Upgrade regex dependency

SpamapS:
There have been a number of performance enhancements which make having
a newer regex desirable. However the most important change was dropping
thread_local in 1.4.4. Thread_local had some security vulnerabilities,
making it trigger lots of dependabot alerts when it lands in Cargo.lock
files.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,30 +4,18 @@ version = 3
 
 [[package]]
 name = "aho-corasick"
-version = "0.7.19"
+version = "1.1.2"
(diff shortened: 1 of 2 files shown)


matthiasbeyer:
As far as I can see, `regex` is not exported from our public interface, but only used in https://github.com/rust-cli/env_logger/blob/main/src/filter/regex.rs :eyes: 

I'll approve CI, but leave the final decision to @epage here.

PR #288, opened 2023-11-20
bump is-terminal

jamesadevine:
is-terminal v0.4.0 introduces potential security hole via rustix. This PR bumps is-terminal to v0.4.9. Read about the security vulnerability here: https://github.com/bytecodealliance/rustix/security/advisories/GHSA-c827-hfw6-qwvm

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -13,15 +13,9 @@ dependencies = [
 
 [[package]]
 name = "bitflags"
-version = "1.3.2"
+version = "2.4.1"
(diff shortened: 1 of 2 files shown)


epage:
I'm not seeing why we should bump our version requirement. This vulnerability doesn't impact us and users of our library can control this through there lockfile.

jamesadevine:
Indeed, turning off the feature upstream is the remediation I have performed. Feel free to close if it does not apply to the codebase and you are not going to accept the version bump :smile:

PR #294, opened 2024-01-09
support kv_unstable

SVilgelm:
Print the kev and values from a record if `kv_unstable` feature enabled


SVilgelm:
I'm aware of the https://github.com/rust-cli/env_logger/pull/137, but that one is really outdated and I decided to write simpler solution

tmccombs:
This is actually pretty similar to what I originally had in #137. But due to some other comments I added support for putting the key/value pairs on multiple lines, and putting it at the end instead of the beginning.

PR #295, opened 2024-01-16
Add `FilteredLog` utility to combine a `Filter` with a `Log`.

de-vri-es:
This can be used by end-users that want to combine the filtering capabilities of `env_logger` with a different `Log` implementation.

In my case, I want to combine the filtering of `env_logger` with `systemd-journal-logger`. I could just do it myself where I need it, but this feels like something that could be useful to provide from `env_logger` already.


de-vri-es:
The `Filter` type is already exposed though. It's true that you could pull in a lot less dependencies if you only want the `Filter` type.

So.. if I make a PR to split `Filter` to a separate crate and re-export it from `env_logger`, would that be a good way forward?

And then would you be inclined to add this utility to wrap an existing logger?

de-vri-es:
Pushed a commit to split the `filter` module off to a separate crate. What do you think?

epage:
I've incorporated parts of this into #299 and #300

PR #302, opened 2024-01-27
fix(fmt): Fix passing of WriteStyle when using Target::Pipe

### LINKED ISSUE #274: `WriteStyle` is ignored when used with `Target::Pipe`

Bobo1239:
Seems like this was overlooked in #298.

Test program:
```Rust
fn main() {
    env_logger::builder()
        .target(env_logger::Target::Pipe(Box::new(std::io::stdout())))
        .write_style(env_logger::WriteStyle::Always)
        .init();
    log::error!("ASD");
}
```

Fixes #274.

### DIFF
--- a/src/fmt/writer/buffer.rs
+++ b/src/fmt/writer/buffer.rs
@@ -31,10 +31,13 @@ impl BufferWriter {
         }
     }
 
-    pub(in crate::fmt::writer) fn pipe(pipe: Box<Mutex<dyn io::Write + Send + 'static>>) -> Self {
+    pub(in crate::fmt::writer) fn pipe(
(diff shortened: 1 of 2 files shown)


epage:
Thanks for that and sorry I missed that!

PR #304, opened 2024-01-31
Workaround cargo bug with enable features of optional dependencies

sdroege:
While this works fine with Rust 1.73+, it does not work with Rust 1.70 and fails like this:

```
error: Package `env_logger v0.11.1` does not have feature `anstream`. It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,7 @@ pre-release-replacements = [
 
 [features]
 default = ["auto-color", "humantime", "regex"]
-color = ["dep:anstream", "dep:anstyle"]
+color = ["anstream", "anstyle"]
 auto-color = ["color", "anstream/auto"]
 humantime = ["dep:humantime"]
 regex = ["env_filter/regex"]


sdroege:
It would be nice to have a bugfix release with this so compilation works with 1.70 again :)

sdroege:
Ah, sorry for missing that.

MSRV 1.71 is fine for me, I'll just stay with 0.10 for the time being until I update this crate of mine to something newer. Only reason why it's still at MSRV 1.70 is because I didn't have the patience yet to deal with the usual people complaining about requiring a "too new" toolchain :)

epage:
I know that pain all too well.

The main reason I was willing to update the MSRV so much in 0.11 is because it had breaking changes associated with it (but I don't bump MSRV for breaking changes).

My hope is the MSRV RFC will reduce that pressure.  As you are affected, if you haven't looked and commented on it yet, I would recommend it!

sdroege:
I looked at one of the initial versions and it seemed like a huge improvement over the current situation. The latest version still looks good to me. Having the resolver consider MSRV would allow me to simply support the latest in most crates and have everything else be the problem of the users with annoying version requirements (-> they would just have to stay at an older version of various crates too then instead of dependency resolution failing).

Thanks for working on this :)

PR #306, opened 2024-02-07
fix(doc): Fix `env_logger` version is outdated in readme doc

rhysd:


README.md
```diff
@@ -82,7 +82,7 @@ Tests can use the `env_logger` crate to see log messages generated during that t
 log = "0.4.0"
 
 [dev-dependencies]
-env_logger = "0.10.0"
+env_logger = "0.11.0"
```
epage, line 85: Instead, could you replace this with `cargo add` commands so we don't have to ensure this is updated

rhysd: I force-pushed the change at [a0ca897](https://github.com/rust-cli/env_logger/pull/306/commits/a0ca897156286414f21d656c1474cd0f9d6c6346).

epage:
Thanks!

PR #324, opened 2024-07-18
Add `Builder::try_parse()` method

### LINKED ISSUE #323: Add a fitler parsing function returning `Result` instead of writing errors to `stderr`
I suggest adding a new method `env_filter::filter::try_parse(&mut self, filters: &str) -> Result<&mut Self, ParsingError>`. This would imply adding a new `ParsingError` type and a small refactoring, but would make the library more convenient in some situations.

In my use case, I don't have access to the contents of `stderr` when the application is starting, and I actually don't want the application to start if the logger is configured incorrectly. Such a method would let me do this without copying the parser implementation to the application code.

If the maintainers agree that this can be added to the `env_logger`, I am happy to create a PR.

Maximkaaa:
Current implementation of the `Builder::parse()` method prints out specification errors to `stderr` and then just ignores them. This is fine for most console applications, but in some cases better control over what is happening is needed:

* Sometimes there is no access to `stderr`, in that case there is no way to find out what went wrong
* For some applications it's more desirable to fail on startup than to run with (partially) invalid configuration.

For such cases this PR introduces a new method `try_parse` that does the same thing as the `parse` method, but returns an `Err` in case an error in the specification is found.

This change was suggested in and fixes #323.

### DIFF
--- a/crates/env_filter/CHANGELOG.md
+++ b/crates/env_filter/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 <!-- next-header -->
 ## [Unreleased] - ReleaseDate
 
+### Features
+
+- Added `env_filter::Builder::try_parse(&self, &str)` method (failable version of `env_filter::Builder::parse()`)
+
 ## [0.1.0] - 2024-01-19
 
 <!-- next-url -->
--- a/crates/env_filter/Cargo.toml
+++ b/crates/env_filter/Cargo.toml
@@ -33,5 +33,8 @@ regex = ["dep:regex"]
 log = { version = "0.4.8", features = ["std"] }
 regex = { version = "1.0.3", optional = true, default-features=false, features=["std", "perf"] }
 
+[dev-dependencies]
+snapbox = "0.6"
+
 [lints]
 workspace = true
--- a/crates/env_filter/src/parser.rs
+++ b/crates/env_filter/src/parser.rs
@@ -87,6 +87,7 @@ pub(crate) fn parse_spec(spec: &str) -> ParseResult {
 #[cfg(test)]
 mod tests {
     use log::LevelFilter;
+    use snapbox::{assert_data_eq, str};
 
     use super::{parse_spec, ParseResult};
 
@@ -95,7 +96,7 @@ mod tests {
         let ParseResult {
             directives: dirs,
             filter,
-            ..
+            errors,
         } = parse_spec("crate1::mod1=error,crate1::mod2,crate2=debug");
 
         assert_eq!(dirs.len(), 3);
@@ -108,6 +109,8 @@ mod tests {
         assert_eq!(dirs[2].name, Some("crate2".to_owned()));
         assert_eq!(dirs[2].level, LevelFilter::Debug);
         assert!(filter.is_none());
+
+        assert!(errors.is_empty());
     }
 
     #[test]
@@ -116,13 +119,19 @@ mod tests {
         let ParseResult {
             directives: dirs,
             filter,
-            ..
+            errors,
         } = parse_spec("crate1::mod1=warn=info,crate2=debug");
 
         assert_eq!(dirs.len(), 1);
         assert_eq!(dirs[0].name, Some("crate2".to_owned()));
         assert_eq!(dirs[0].level, LevelFilter::Debug);
         assert!(filter.is_none());
+
+        assert_eq!(errors.len(), 1);
+        assert_data_eq!(
+            &errors[0],
+            str!["invalid logging spec 'crate1::mod1=warn=info'"]
+        );
     }
 
     #[test]
@@ -131,13 +140,16 @@ mod tests {
         let ParseResult {
             directives: dirs,
             filter,
-            ..
+            errors,
         } = parse_spec("crate1::mod1=noNumber,crate2=debug");
 
         assert_eq!(dirs.len(), 1);
         assert_eq!(dirs[0].name, Some("crate2".to_owned()));
         assert_eq!(dirs[0].level, LevelFilter::Debug);
         assert!(filter.is_none());
+
+        assert_eq!(errors.len(), 1);
+        assert_data_eq!(&errors[0], str!["invalid logging spec 'noNumber'"]);
     }
 
     #[test]
@@ -146,13 +158,16 @@ mod tests {
         let ParseResult {
             directives: dirs,
             filter,
-            ..
+            errors,
         } = parse_spec("crate1::mod1=wrong,crate2=warn");
 
         assert_eq!(dirs.len(), 1);
         assert_eq!(dirs[0].name, Some("crate2".to_owned()));
         assert_eq!(dirs[0].level, LevelFilter::Warn);
         assert!(filter.is_none());
+
+        assert_eq!(errors.len(), 1);
+        assert_data_eq!(&errors[0], str!["invalid logging spec 'wrong'"]);
     }
 
     #[test]
@@ -161,13 +176,16 @@ mod tests {
         let ParseResult {
             directives: dirs,
             filter,
-            ..
+            errors,
         } = parse_spec("crate1::mod1=wrong,crate2=");
 
         assert_eq!(dirs.len(), 1);
         assert_eq!(dirs[0].name, Some("crate2".to_owned()));
         assert_eq!(dirs[0].level, LevelFilter::max());
         assert!(filter.is_none());
+
+        assert_eq!(errors.len(), 1);
+        assert_data_eq!(&errors[0], str!["invalid logging spec 'wrong'"]);
     }
 
     #[test]
@@ -176,10 +194,11 @@ mod tests {
         let ParseResult {
             directives: dirs,
             filter,
-            ..
+            errors,
         } = parse_spec(""); // should be ignored
         assert_eq!(dirs.len(), 0);
         assert!(filter.is_none());
+        assert!(errors.is_empty());
     }
 
     #[test]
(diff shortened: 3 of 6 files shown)


crates/env_filter/src/parser.rs
```diff
@@ -42,20 +77,26 @@ pub(crate) fn parse_spec(spec: &str) -> (Vec<Directive>, Option<FilterOp>) {
...
+                                    part1
+                                ),
+                            )?;
                             continue;
                         }
                     }
                     _ => {
-                        eprintln!(
-                            "warning: invalid logging spec '{}', \
+                        err_or_print(
```
epage, line 92: Not too thrilled with `err_or_print`

I'm hoping we can instead have the following commits
- refactor commit to change `parse_spec` to return an error with `Builder::parse` taking the error and printing it as a warning
- feature commit that pulls the code out of `Builder::parse` into `Builder::try_parse`, making `Builder::parse` a wrapper around it that takes the error and prints it as a warning

crates/env_filter/src/parser.rs
```diff
@@ -1,25 +1,60 @@
 use log::LevelFilter;
+use std::error::Error;
+use std::fmt::{Debug, Display, Formatter};
 
 use crate::Directive;
 use crate::FilterOp;
 
+/// Error during logger directive parsing process.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ParsingError {
```
epage, line 10: nit: Feel like this should be named `ParseError` rather than `ParsingError`

crates/env_filter/src/parser.rs
```diff
@@ -1,25 +1,60 @@
...
+    /// Details of what exactly was wrong in the directive.
+    pub details: String,
```
epage, line 12: The fields should not be public

crates/env_filter/src/parser.rs
```diff
@@ -66,22 +107,39 @@ pub(crate) fn parse_spec(spec: &str) -> (Vec<Directive>, Option<FilterOp>) {
...
+        .map(|filter| match FilterOp::new(filter) {
+            Ok(re) => Ok(Some(re)),
+            Err(e) => err_or_print(
+                skip_errors,
+                format!("warning: invalid regex filter - {}", e),
+            )
+            .map(|_| None),
+        })
+        .transpose()?
+        .flatten();
```
epage, line 120: This is getting to be too complex of combinators.  I'm assuming there is a more straightforward imperative way of writing this.  Maybe refactoring the original code could also be its own commit, paving the way for the added complexity of this

crates/env_filter/src/parser.rs
```diff
@@ -108,6 +166,11 @@ mod tests {
         assert!(filter.is_none());
     }
 
+    #[test]
+    fn try_parse_spec_invalid_crate() {
+        assert!(try_parse_spec("crate1::mod1=warn=info,crate2=debug", false).is_err());
+    }
```
epage, line 172: Let's add `snapbox` as a dev dependency and then you can do
```rust
let input = "crate1::mod1=warn=info,crate2=debug";
let actual = try_parse_spec(input).unwrap_err();
assert_data_eq!(actual, str![]);
```
and then use snapshot testing to fill in the expected result.

This will ensure we test the quality of the messages for a given string *and* can easily update the tests if we work to improve the quality of the messages

crates/env_filter/src/parser.rs
Maximkaaa: I agree, this was not the best solution. But `parse_spec` cannot just return `Result`, as `Builder::parse` method needs the valid part of the spec even in case of an error. Also, it is possible to have multiple errors in the same spec, so just bailing out with the first error would also not work.

Instead, I've added a new `ParseResult` struct, that contains original returned values plus the encountered errors, and made the caller decide, what they want to do with these errors. I believe this approach looks quite clean now.

crates/env_filter/src/parser.rs
Maximkaaa: Done

crates/env_filter/src/parser.rs
Maximkaaa: Done

crates/env_filter/src/parser.rs
Maximkaaa: With the new `ParseResult` this code seems much cleaner.

crates/env_filter/src/parser.rs
Maximkaaa: Thanks for the pointers to the `snapbox`. It's really convenient!

crates/env_filter/Cargo.toml
```diff
@@ -18,11 +18,11 @@ rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.release]
 pre-release-replacements = [
-  {file="CHANGELOG.md", search="Unreleased", replace="{{version}}", min=1},
```
Maximkaaa, line 21: Sorry, missed the `toml` file being autoformatted... Shall I try to revert this?

crates/env_filter/src/parser.rs
```diff
@@ -3,23 +3,42 @@ use log::LevelFilter;
...
+pub(crate) fn parse_spec(spec: &str) -> ParseResult {
+    let mut result = ParseResult::default();
 
     let mut parts = spec.split('/');
     let mods = parts.next();
     let filter = parts.next();
     if parts.next().is_some() {
-        eprintln!(
+        result.add_error(format!(
             "warning: invalid logging spec '{}', \
```
epage: The `warning:` prefix should be in the caller

crates/env_filter/Cargo.toml
epage: Please do so as its unrelated to tis commit

crates/env_filter/src/parser.rs
epage: Ah, you did that in the follow up but I  feel like that should belong in this commit

crates/env_filter/src/parser.rs
```diff
@@ -22,8 +24,36 @@ impl ParseResult {
...
+            Err(ParseError {
+                details: self.errors,
+            })
+        }
+    }
 }
 
+/// Error during logger directive parsing process.
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct ParseError {
```
epage, line 41: An aliases commit makes reference to `ParseError`

crates/env_filter/src/parser.rs
```diff
@@ -22,8 +24,36 @@ impl ParseResult {
     fn add_error(&mut self, message: String) {
         self.errors.push(message);
     }
+
+    pub(crate) fn ok(self) -> Result<(Vec<Directive>, Option<FilterOp>), ParseError> {
+        if self.errors.is_empty() {
+            Ok((self.directives, self.filter))
+        } else {
+            Err(ParseError {
+                details: self.errors,
```
epage, line 33: Unsure how I feel about printing the details comma separated.  Should we just return the first?

epage:
Thanks for moving this forward!

crates/env_filter/src/parser.rs
Maximkaaa: I would argue that the more details an error message provides, the better. In this case, we already allocate for `String` message in the error, so giving more details is doesn't have downsides in my opinion. It's often frustrating when a system returns only first error of many when it easily can save you several iterations of fixing your code by simply telling you all your errors right away.

If you find `Vec` in the error struct inappropriate, we could join strings here instead of doing it in the `Display` (although I think it's better to keep messages separated in the error struct).

crates/env_filter/Cargo.toml
Maximkaaa: Done

crates/env_filter/src/parser.rs
Maximkaaa: Moved it into the first commit

crates/env_filter/src/parser.rs
Maximkaaa: Fixed

crates/env_filter/src/parser.rs
```diff
@@ -98,17 +86,26 @@ pub(crate) fn parse_spec(spec: &str) -> ParseResult {
...
 mod tests {
+    use crate::ParseError;
     use log::LevelFilter;
+    use snapbox::{assert_data_eq, str, Data, IntoData};
 
     use super::{parse_spec, ParseResult};
 
+    impl IntoData for ParseError {
+        fn into_data(self) -> Data {
+            format!("{self}").into_data()
```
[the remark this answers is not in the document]
Maximkaaa: Done

Maximkaaa:
I fixed all the comments except the last one, since I think that the current approach is better. If you insist on returning just one error string, I will edit it also.

crates/env_filter/CHANGELOG.md
```diff
@@ -1,14 +1,22 @@
 # Change Log
+
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
 <!-- next-header -->
+
 ## [Unreleased] - ReleaseDate
```
epage, line 11: Please revert these unrelated changes

crates/env_filter/src/parser.rs
epage: I'm not worried about performance but the usability of concatenating the errors. imo we should only return one.  If someone finds a concrete use case where this is inhibiting usability enough, let's evaluate a "error recovery" API that returns a `Vec<ParseError>` (or some other related approach).

crates/env_filter/CHANGELOG.md
Maximkaaa: Sorry for that. Done.

crates/env_filter/src/parser.rs
Maximkaaa: All right, I've made the error contain the first error message.

epage:
Looks like the final commit needs `cargo fmt` run on it

Maximkaaa:
> Looks like the final commit needs `cargo fmt` run on it

Done.

epage:
Thanks!  This is now released.

PR #328, opened 2024-07-27
Update example code to run inside `main`

keoghpe:
Small docs change - I'm new to this crate and found the [current docs page](https://docs.rs/env_logger/latest/env_logger/) confusing as it calls init outside a `main` function. Hopefully this makes it clearer!

Thanks for creating a cool crate ❤️


epage:
I'm mixed on this.  This makes the intent even more explicit.  On the other hand, its not very idiomatic these days to have `main` in doc tests.

Issue #337, opened 2024-10-08
anstream feature does not exist


anacrolix:
```
> cargo +nightly udeps
error: failed to select a version for `env_logger`.
    ... required by package `replica v1.0.4 (/Users/anacrolix/getlantern/replica-rust)`
versions that meet the requirements `*` (locked to 0.11.5) are: 0.11.5

the package `replica` depends on `env_logger`, with features: `anstream` but `env_logger` does not have these features.
 It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.


failed to select a version for `env_logger` which could resolve this conflict
```

epage:
Sounds like a bug in replica as that feature doesn't exist.

anacrolix:
Ah no it's because the default feature lists anstream as a default, but anstream is a feature of a dependency.

epage:
As `env_logger` doesn't have an `anstream` feature (its an optional dependency enabled via `color`), this seems to be a bug in replica-rust and not env_logger, so I'm going to go ahead and close this.

Issue #312, opened 2024-02-18
Issue with `anstream` optional dependency


vantessel:
Running `cargo add log env_logger` in a fresh crate gives me this error:
```
error: failed to select a version for `env_logger`.
    ... required by package `testlog v0.1.0`
versions that meet the requirements `^0.11.2` are: 0.11.2

the package `testlog` depends on `env_logger`, with features: `anstream` but `env_logger` does not have these features.
 It has an optional dependency with that name, but that dependency uses the "dep:" syntax in the features table, so it does not have an implicit feature with that name.


failed to select a version for `env_logger` which could resolve this conflict
```

It works fine with version `0.10.2`

epage:
FYI v0.11's MSRV is 1.71 so we could rely on a fix in cargo related to feature declarations.  Unfortunately, cargo reports the error before checking the MSRV.

This previously came up in #304.

PR #338, opened 2024-10-10
Fix introduced dependency on anstream by auto-color

anacrolix:
Fixes https://github.com/rust-cli/env_logger/issues/337.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -118,7 +118,7 @@ pre-release-replacements = [
 [features]
 default = ["auto-color", "humantime", "regex"]
 color = ["dep:anstream", "dep:anstyle"]
-auto-color = ["color", "anstream/auto"]
+auto-color = ["color", "anstream?/auto"]
 humantime = ["dep:humantime"]
 regex = ["env_filter/regex"]
 unstable-kv = ["log/kv"]


epage:
I'm still not seeing what the problem is here.

anacrolix:
@vantessel @sdroege might have some insight. 

Related:

https://github.com/rust-cli/env_logger/issues/312

https://github.com/rust-cli/env_logger/pull/304

sdroege:
Yes, at this point this doesn't seem worth it anymore

anacrolix:
auto-color creates a dependency on anstream when one should not exist. It confuses the tooling. I'm using nightly.

epage:
What tooling does it confuse? Our CI passes and many other people use it.

anacrolix:
https://github.com/rust-cli/env_logger/issues/337#issue-2574392620

epage:
As its working everywhere else in Cargo, I'm assuming that is a bug in `cargo-udeps`.  As such, it should be addressed there.  Before opening an issue there, I would recommend making sure your install is up to date in case that is the issue.  They link against cargo and so if you are using a version with an old enough cargo, then yes it would be broken.

bburdette:
Getting this message today, with cargo 1.77.1

PR #351, opened 2025-03-08
Replace humantime with cyborgtime

oherrala:
[`humantime`](https://crates.io/crates/humantime/) seems to be unmaintained and [`cyborgtime`](https://crates.io/crates/cyborgtime/) is fork of it. This change shouldn't affect end users. E.g. "humantime" feature is kept as is.

I also introduced "cyborgtime" feature that currently works same as "humantime".

Ref. https://github.com/rustsec/advisory-db/pull/2249

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,6 +66,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "acbf1af155f9b9ef647e42cdc158db4b64a1b61f743629225fde6f3e0be2a7c7"
 
 [[package]]
+name = "cyborgtime"
+version = "2.1.1"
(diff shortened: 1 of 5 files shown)


BurntSushi:
(Disclosure: I'm the author of Jiff.)

This kinda looks like you're replacing one unmaintained dependency with another. `cyborgtime` was forked ~2 years ago, had a few commits and then hasn't seen any activity since. It doesn't seem like the right play here?

On the other hand, bringing in something like Jiff (or even chrono or time) seems a little heavyweight for this. I'm not sure there is really much of a middle ground other than sticking with `humantime`. Then again, I think this is optional (albeit enabled by default), so maybe it's fine.

oherrala:
@epage @BurntSushi I made a `jiff` version in #352.

PR #352, opened 2025-03-08
chore(deps): Switch humantime to jiff

oherrala:
[`humantime`](https://crates.io/crates/humantime/) seems to be unmaintained and [`jiff`](https://crates.io/crates/jiff/) provides the same functionality. This change shouldn't affect end users.

Ref. https://github.com/rustsec/advisory-db/pull/2249

### DIFF
--- a/src/fmt/humantime.rs
+++ b/src/fmt/humantime.rs
@@ -109,3 +109,28 @@ impl fmt::Display for Timestamp {
         formatter(self.time).fmt(f)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::Timestamp;
+    use crate::TimestampPrecision;
+
+    #[test]
+    fn test_display_timestamp() {
+        let mut ts = Timestamp {
+            time: std::time::SystemTime::UNIX_EPOCH,
+            precision: TimestampPrecision::Nanos,
+        };
+
+        assert_eq!("1970-01-01T00:00:00.000000000Z", format!("{ts}"));
+
+        ts.precision = TimestampPrecision::Micros;
+        assert_eq!("1970-01-01T00:00:00.000000Z", format!("{ts}"));
+
+        ts.precision = TimestampPrecision::Millis;
+        assert_eq!("1970-01-01T00:00:00.000Z", format!("{ts}"));
+
+        ts.precision = TimestampPrecision::Seconds;
+        assert_eq!("1970-01-01T00:00:00Z", format!("{ts}"));
+    }
+}
--- a/src/fmt/humantime.rs
+++ b/src/fmt/humantime.rs
@@ -1,10 +1,6 @@
 use std::fmt;
 use std::time::SystemTime;
 
-use humantime::{
-    format_rfc3339_micros, format_rfc3339_millis, format_rfc3339_nanos, format_rfc3339_seconds,
-};
-
 use crate::fmt::{Formatter, TimestampPrecision};
 
 impl Formatter {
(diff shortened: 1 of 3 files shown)


src/fmt/humantime.rs
```diff
@@ -99,13 +95,46 @@ impl fmt::Debug for Timestamp {
...
         let formatter = match self.precision {
-            TimestampPrecision::Seconds => format_rfc3339_seconds,
-            TimestampPrecision::Millis => format_rfc3339_millis,
-            TimestampPrecision::Micros => format_rfc3339_micros,
-            TimestampPrecision::Nanos => format_rfc3339_nanos,
+            TimestampPrecision::Seconds => |t| strtime::format("%FT%TZ", t),
+            TimestampPrecision::Millis => |t| strtime::format("%FT%T%.3fZ", t),
+            TimestampPrecision::Micros => |t| strtime::format("%FT%T%.6fZ", t),
+            TimestampPrecision::Nanos => |t| strtime::format("%FT%T%.9fZ", t),
+        };
```
BurntSushi, line 109: It's not necessary to use strftime for this. It's likely slower. The `Timestamp` display impl will use rfc 3339 already.

And you can control precision using Rust's standard formatting facilities: https://docs.rs/jiff/latest/jiff/struct.Timestamp.html#example-35

oherrala: Thanks! That was an awesome tip. Fixed!

src/fmt/humantime.rs
```diff
@@ -99,13 +95,40 @@ impl fmt::Debug for Timestamp {
...
+        ts.precision = TimestampPrecision::Micros;
+        assert_eq!("1970-01-01T00:00:00.000000Z", format!("{ts}"));
+
+        ts.precision = TimestampPrecision::Millis;
+        assert_eq!("1970-01-01T00:00:00.000Z", format!("{ts}"));
+
+        ts.precision = TimestampPrecision::Seconds;
+        assert_eq!("1970-01-01T00:00:00Z", format!("{ts}"));
     }
 }
```
epage, line 134: Can you add the tests (passing) in their own commit before the implementation change?  That way, this commit shows there is no behavior change (or shows what it is and we can sign off on it)

oherrala: @epage Good idea. Implemented!

epage:
Thanks!

djc:
Note that the addition of jiff means env_logger is pulling a number of extra dependencies: portable-atomic, portable-atomic-util, serde (including serde_derive and all the proc-macro stuff). Not sure this is a good deal from that perspective.

(Independent from my having taken over maintenance of humantime.)

yhx-12243:
Yes I agree, I have a little intention to propose a revert to this PR.

BurntSushi:
> Note that the addition of jiff means env_logger is pulling a number of extra dependencies: portable-atomic, portable-atomic-util, serde (including serde_derive and all the proc-macro stuff). Not sure this is a good deal from that perspective.

They might show up in your lock file because of https://github.com/rust-lang/cargo/issues/10801, but serde, proc macros are all opt-in features that [`env_logger` is not enabling](https://github.com/rust-lang/cargo/issues/10801). They should not being built without opting into the relevant features, and if they are, you're more than welcome to file a Jiff bug with an MRE. And `portable-atomic` is only going to be used on [targets that don't have atomics](https://github.com/BurntSushi/jiff/blob/8d9ff390202399af4dbac326e127fd34291fb2ae/Cargo.toml#L229).

djc:
@BurntSushi good to know! We should really get that Cargo bug fixed sometime...

PR #356, opened 2025-03-28
Added example of timer

vigna:


src/lib.rs
```diff
@@ -256,6 +256,22 @@
...
+//! use std::io::Write;
+//!
+//! let start = std::time::Instant::now();
+//! env_logger::builder()
+//!     .format(move |buf, record| {
+//!         writeln!(buf, "{} ({}): {}", record.level(), start.elapsed().as_millis(), record.args())
+//!     })
+//!     .init();
+//! ```
+//!
```
epage, line 274: I was assuming this would be created as another example in `examples/`.  The advantage of this is its easier to interactively play with to see the results.

If its being added to the docs, putting it on [`Builder::format`](https://docs.rs/env_logger/latest/env_logger/struct.Builder.html#method.format) might be a better approach.

vigna: So sorry to disappoint your expectations 😆.

PR #358, opened 2025-03-30
feat: Syslog-adapted formating and auto-detection

### LINKED ISSUE #349: Systemd log format activable via env var
I updated my program with something along the lines of https://github.com/rust-cli/env_logger/blob/main/examples/syslog_friendly_format.rs to have a systemd friendly logging, however when I use other binaries using env_logger usually the same logic is not available, can we have it enable-able by default in env_logger?

ie if `RUST_LOG_STYLE=SYSTEMD` the default formatting is changed with the systemd friendly one without having to change the downstream lib

### LINKED ISSUE #357: better support for logging to syslog
this is kinda like #349 but bigger.

Have two format strings, the regular one is configured just as it currently is, and a second one that could be configured via .format_syslog in the builder.

The syslog formatter by default skips timestamps and such as those will be provided by the syslog.

env_logger then auto-detects if its running as a daemon, with an override available.
The most common service manager is systemd.
Detecting being run as systemd daemon is as simple as checking for the INVOCATION_ID environment variable.
Other service managers may require more complex detection.

I am willing to contribute the code, in fact i am working on it already, and will send a pr later.

djugei:
closes #357
closes #349 

this is a surprisingly low diff size.

### DIFF
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -274,7 +274,7 @@ impl<T: Display> Display for StyledValue<T> {
 #[cfg(not(feature = "color"))]
 type StyledValue<T> = T;
 
-struct SyslogFormatter;
+pub(crate) struct SyslogFormatter;
 
 impl SyslogFormatter {
     pub(crate) fn build(&mut self) -> FormatFn {
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -63,7 +63,6 @@ use std::io::prelude::Write;
 use std::rc::Rc;
 use std::{fmt, io, mem};
 
-#[cfg(feature = "color")]
 use log::Level;
 use log::Record;
 
@@ -274,6 +273,28 @@ impl<T: Display> Display for StyledValue<T> {
 
 #[cfg(not(feature = "color"))]
 type StyledValue<T> = T;
+
+struct SyslogFormatter;
+
+impl SyslogFormatter {
+    pub(crate) fn build(&mut self) -> FormatFn {
+        Box::new(|buf: &mut Formatter, record: &Record<'_>| {
+            writeln!(
+                buf,
+                "<{}>{}: {}",
+                match record.level() {
+                    Level::Error => 3,
+                    Level::Warn => 4,
+                    Level::Info => 6,
+                    Level::Debug => 7,
+                    Level::Trace => 7,
+                },
+                record.target(),
+                record.args()
+            )
+        })
+    }
+}
 
 /// A [custom format][crate::Builder::format] with settings for which fields to show
 pub struct ConfigurableFormat {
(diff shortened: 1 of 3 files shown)


epage:
We do not have agreement that this solution is in the direction we want to go and so this is only a proposal for demonstrating a direction, regardless of how useful it is to you in the moment.  Having this be a draft (or closed) clarifies how this is being treated and pushes discussion on solutions to a single place, the issue.

djugei:
oh, ok so a draft from the project side, i guess that makes sense. i usually see the draft mode used from the contributor side to indicate work-in-progress status.

djugei:
i kinda want to slightly alter the log format, right now syslog contains

`date host processname[process_id] processname::module:: message`

i would like
`date host processname[process_id] ::module:: message`
better to reduce redundancy.

i am unsure what the easiest way to get the top level binary name is though. then it would be a simple .strip_prefix(name).

djugei:
should be merged after this: https://github.com/rust-cli/env_logger/pull/363

djugei:
rebased

djugei:
i have rebased this to the latest main branch, i have also rebased it to not require #363 anymore.

src/logger.rs
```diff
@@ -336,6 +341,13 @@ impl Builder {
         self
     }
 
+    /// If set to true, format log messages in a Syslog-adapted format.
+    /// Overrides the auto-detected value.
```
epage, line 346: newline to break the paragraphs up

src/logger.rs
```diff
@@ -820,6 +836,11 @@ impl<'a> Env<'a> {
     fn get_write_style(&self) -> Option<String> {
         self.write_style.get()
     }
+
+    fn is_daemon(&self) -> bool {
+        //TODO: support more logging systems
```
epage: Either
- we are actively tracking this TODO and so it should be in our backlog (issues)
- we are not actively tracking this in which case this is more of a comment that isn't really clarifying things

Note also that other logging systems would likely have other formats, so likely the name of this function is wrong or this really needs to emit an `Option` of an enum.

src/fmt/mod.rs
```diff
@@ -275,6 +274,28 @@ impl<T: Display> Display for StyledValue<T> {
 #[cfg(not(feature = "color"))]
 type StyledValue<T> = T;
 
+pub struct SyslogFormatter;
+
+impl SyslogFormatter {
```
epage: This is `pub` but this isn't usable by the caller

src/logger.rs
djugei: i have changed the comment, the thing about being Option<Format> is a good point, though as this is an internal api we can just adapt it as/if more formats get added, to the best of my knowledge the current format is correct for about any unix system though, and all issues i have seen on here were about unix systems.

sant0s12:
Imo this approach is a bit restrictive since it ties the priority prefixes to the format. It would be nice to be able to customize the format while enabling/disabling the prefixes.

PR #360, opened 2025-04-01
feat(fmt): Expose ConfigurableFormat

epage:
The goal is to allow people to compose more complex formatting options (e.g. for #349) by exposing the built-in formatting

A breaking change was split out into #361

### DIFF
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -276,10 +276,9 @@ impl<T: Display> Display for StyledValue<T> {
 #[cfg(not(feature = "color"))]
 type StyledValue<T> = T;
 
-/// The default format.
-///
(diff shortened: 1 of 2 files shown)


djugei:
i do not think the builder which modifies defaultformat would ever be useful for any user of the library that wants a custom formatter. no functions interact at all. i think those two should be fully decoupled.

epage:
> this seems somewhat redundant, customformat or default format is used, i do not quite understand why both exist at the same time.
tbh i don't fully understand the intent behind this entire pr.

This PR serves the following roles
- Makes the intent clearer in what is associated with the default formatter and not the custom formatter
- Allows callers to create their own formatters on top of `ConfigurableFormat`

PR #361, opened 2025-04-01
feat(fmt)!: Generalize custom formatter to 'RecordFormat' 

epage:
This breaks type inference for basic custom formatters


djugei:
I have tested a similar change for #363 and i recommend against it.

Too much type info annoyance. I would instead recommend having two separate functions, one for the common case (FormatFn) and one for special needs (RecordFormat)

PR #363, opened 2025-04-19
refactor: move format/with_format from builder to formatter

djugei:
as mentioned in #360


djugei:
obviously a breaking change cause the interface is different.

still unsure if that signature should be used (more generic but requires type annotations) or if what was on builder should be kept (less generic, no type annotations) or maybe both with different names.

currently leaning towards both.

PR #378, opened 2025-08-22
feat: Adding support for `no_std` environments for `env_filter`

WolverinDEV:
Using `env_logger` in a `no_std` environment may not seem useful at first. However, the `env_filter` crate is very convenient for defining complex filters that can then be implemented by a custom `no_std`-compatible logger.  

This approach leverages the convenience developers are already familiar with from `env_logger`, while enabling the same flexibility in `no_std` contexts with other logging backends.

### DIFF
--- a/crates/env_filter/Cargo.toml
+++ b/crates/env_filter/Cargo.toml
@@ -31,8 +31,8 @@ regex = ["std", "dep:regex"]
 std = []
 
 [dependencies]
-log = { version = "0.4.8", features = ["std"] }
-regex = { version = "1.0.3", optional = true, default-features=false, features=["std", "perf"] }
+log = { version = "0.4.8", default-features = false }
+regex = { version = "1.0.3", optional = true, default-features = false, features = ["std", "perf"] }
 
 [dev-dependencies]
 snapbox = "0.6"


crates/env_filter/src/parser.rs
```diff
@@ -46,12 +50,16 @@ pub struct ParseError {
...
         write!(f, "error parsing logger filter: {}", self.details)
     }
 }
 
-impl Error for ParseError {}
+#[cfg(feature = "std")]
+impl std::error::Error for ParseError {}
+
+#[cfg(not(feature = "std"))]
+impl core::error::Error for ParseError {}
```
epage, line 62: This makes the lack of `std` require a higher MSRV

crates/env_filter/src/parser.rs
```diff
@@ -1,6 +1,10 @@
...
+use alloc::string::String;
+use alloc::vec::Vec;
+
+use core::fmt::{Display, Formatter};
 use log::LevelFilter;
-use std::error::Error;
-use std::fmt::{Display, Formatter};
 
 use crate::Directive;
 use crate::FilterOp;
```
epage, line 10: If we're changing grouping, could you do
```
use {core,alloc,std}

use third-party

use crate
```

crates/env_filter/Cargo.toml
epage, line 1: > refactor(filter): Add support for no_std environments 

Mind fixing your commit type?

A refactor, by definition, has no user facing changes while this does.  I would use `feat`

PR #404, opened 2026-04-02
feat(env_filter): support no_std

cagatay-y:
Based on #378, this PR adds a new default feature `std` that allows the crate to be used in no_std crates. I tried to incorporate the feedback that was given to the previous PR. Hopefully, I didn't miss anything.

The PR is still a breaking change for crates that have the default features disabled and use `Builder::from_env`, or expect `Builder::parse` to print warnings to stderr and don't have the log target set to stderr.

It may make sense to squash 9c70eafd442e7ec33a6e04cdebd3f50a77af444b and 663c9c22452bb588c4c8e6ee187e659f10132092 (and potentially b754c66235c7ef0e8af03954b2816e93f9fdb709). I kept them separate to make sure I don't misattribute the work of the author of the base PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,9 @@ jobs:
     - uses: Swatinem/rust-cache@v2
     - uses: taiki-e/install-action@cargo-hack
     - name: Default features
-      run: cargo hack check --each-feature --locked --rust-version --ignore-private --workspace --all-targets --keep-going
+      run: |
+        cargo hack check --each-feature --locked --rust-version --ignore-private --package env_logger --all-targets --keep-going
+        cargo hack check --each-feature --locked --version-range 1.81..=1.81 --ignore-private --package env_filter --all-targets --keep-going
   minimal-versions:
     name: Minimal versions
     strategy:
--- a/crates/env_filter/Cargo.toml
+++ b/crates/env_filter/Cargo.toml
@@ -9,7 +9,6 @@ keywords = ["logging", "log", "logger"]
 repository.workspace = true
 license.workspace = true
 edition.workspace = true
-rust-version.workspace = true
 include.workspace = true
 
 [package.metadata.docs.rs]
@@ -26,12 +25,13 @@ pre-release-replacements = [
 ]
 
 [features]
-default = ["regex"]
+default = ["std", "regex"]
 regex = ["dep:regex"]
+std = ["regex/std"]
 
 [dependencies]
-log = { version = "0.4.29", features = ["std"] }
-regex = { version = "1.12.3", optional = true, default-features=false, features=["std", "perf"] }
+log = { version = "0.4.29", default-features = false }
+regex = { version = "1.12.3", optional = true, default-features=false, features=["perf"] }
 
 [dev-dependencies]
 snapbox = "1.0"
--- a/crates/env_filter/Cargo.toml
+++ b/crates/env_filter/Cargo.toml
@@ -26,12 +26,13 @@ pre-release-replacements = [
 ]
 
 [features]
-default = ["regex"]
+default = ["std", "regex"]
 regex = ["dep:regex"]
+std = ["log/std", "regex?/std"]
 
 [dependencies]
-log = { version = "0.4.29", features = ["std"] }
-regex = { version = "1.12.3", optional = true, default-features=false, features=["std", "perf"] }
+log = { version = "0.4.29" }
+regex = { version = "1.12.3", optional = true, default-features=false, features=["perf"] }
 
 [dev-dependencies]
 snapbox = "1.0"
--- a/crates/env_filter/src/filter.rs
+++ b/crates/env_filter/src/filter.rs
@@ -1,6 +1,5 @@
-use std::env;
-use std::fmt;
-use std::mem;
+use alloc::{borrow::ToOwned, string::ToString, vec::Vec};
+use core::{fmt, mem};
 
 use log::{LevelFilter, Metadata, Record};
 
@@ -48,10 +47,11 @@ impl Builder {
     }
 
     /// Initializes the filter builder from an environment.
+    #[cfg(feature = "std")]
     pub fn from_env(env: &str) -> Builder {
         let mut builder = Builder::new();
 
-        if let Ok(s) = env::var(env) {
+        if let Ok(s) = std::env::var(env) {
             builder.parse(&s);
         }
 
@@ -108,7 +108,10 @@ impl Builder {
         } = parse_spec(filters);
 
         for error in errors {
+            #[cfg(feature = "std")]
             eprintln!("warning: {error}, ignoring it");
+            #[cfg(not(feature = "std"))]
+            log::warn!("{error}, ignoring it");
         }
 
         self.filter = filter;
@@ -258,8 +261,9 @@ impl fmt::Debug for Filter {
 
 #[cfg(test)]
 mod tests {
+    use alloc::{borrow::ToOwned, vec, vec::Vec};
+
     use log::{Level, LevelFilter};
-    use snapbox::{assert_data_eq, str};
 
     use super::{enabled, Builder, Directive, Filter};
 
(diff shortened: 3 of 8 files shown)


crates/env_filter/src/lib.rs
```diff
@@ -38,10 +38,13 @@
 //! ```
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(not(feature = "std"), no_std)]
 #![warn(missing_docs)]
 #![warn(clippy::print_stderr)]
 #![warn(clippy::print_stdout)]
```
epage, line 44: For a no-std crate, I would expect the following lints added
```
#![warn(clippy::std_instead_of_core)]
#![warn(clippy::std_instead_of_alloc)]
```

.github/workflows/ci.yml
```diff
@@ -72,7 +72,13 @@ jobs:
...
     - uses: taiki-e/install-action@cargo-hack
     - name: Default features
-      run: cargo hack check --each-feature --locked --rust-version --ignore-private --workspace --all-targets --keep-going
+      run:
+        - cargo hack check --each-feature --locked --rust-version --ignore-private --package env_logger --all-targets --keep-going
+        # When the std feature is enabled, a lower MSRV of 1.71 is sufficient and was the MSRV for the crate before std was made optional.
+        # To make sure we keep it that way, check configurations with std enabled and disabled separately and with their respective MSRVs.
+        # "--at-least-one-of" requires "--feature-powerset", which can be made to emulate "--each-feature" with "--depth 1".
+        - cargo hack check --feature-powerset --depth 1 --locked --version-range 1.71..=1.71 --ignore-private --package env_filter --all-targets --keep-going --at-least-one-of std,default,regex
+        - cargo hack check --no-default-features --locked --version-range 1.81..=1.81 --ignore-private --package env_filter --all-targets --keep-going
```
epage: 1. This is too complicated
2. I am not keen on having feature-dependent MSRVs

epage:
FYI I expect PRs to be presented for how they should be reviewed and merged and not for how they were developed.  Please clean up the commit history

crates/env_filter/src/parser.rs
```diff
@@ -46,12 +47,17 @@ pub struct ParseError {
...
 impl Display for ParseError {
-    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+    fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
         write!(f, "error parsing logger filter: {}", self.details)
     }
 }
 
-impl Error for ParseError {}
+#[cfg(feature = "std")]
+#[allow(clippy::std_instead_of_core)]
```
cagatay-y, line 56: This allow stopped working starting from Rust 1.91. I could not find anything in the release notes that would affect it.

cagatay-y:
@epage, could you take another look at the PR?

.github/workflows/ci.yml
```diff
@@ -72,7 +72,9 @@ jobs:
     - uses: Swatinem/rust-cache@v2
     - uses: taiki-e/install-action@cargo-hack
     - name: Default features
-      run: cargo hack check --each-feature --locked --rust-version --ignore-private --workspace --all-targets --keep-going
+      run: |
+        cargo hack check --each-feature --locked --rust-version --ignore-private --package env_logger --all-targets --keep-going
+        cargo hack check --each-feature --locked --version-range 1.81..=1.81 --ignore-private --package env_filter --all-targets --keep-going
```
epage: Why is this needed?

crates/env_filter/src/filter.rs
```diff
@@ -108,7 +108,10 @@ impl Builder {
         } = parse_spec(filters);
 
         for error in errors {
+            #[cfg(feature = "std")]
             eprintln!("warning: {error}, ignoring it");
+            #[cfg(not(feature = "std"))]
+            log::warn!("{error}, ignoring it");
```
epage: This is for creating the logger in the first place, so I don't think it is too meaningful to log at this point.

crates/env_filter/src/lib.rs
```diff
@@ -38,9 +38,14 @@
 //! ```
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(not(feature = "std"), no_std)]
```
epage: In  my packages I tend to do
```rust
#![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
```
to avoid needing to do weird stuff in the tests

crates/env_filter/Cargo.toml
```diff
@@ -9,7 +9,6 @@ keywords = ["logging", "log", "logger"]
 repository.workspace = true
 license.workspace = true
 edition.workspace = true
-rust-version.workspace = true
```
epage, line 12: Why does this now have no MSRV?

cagatay-y: `error::Error` under `core`, which is used when the `std` feature is not enabled, only got stabilised in Rust 1.81, so the MSRV check fails with the workspace MSRV of 1.71. On the other hand, providing a `env_filter` MSRV of 1.81 causes the `env_logger` MSRV check to fail, even though it enables the `std` feature of `env_filter` and thus is not affected by the absence of `core::error::Error`. I thought that increasing the MSRV for both to 1.81 would unnecessarily negatively affect the users of `env_logger`.

.github/workflows/ci.yml
cagatay-y: Even though it is not possible to declare a MSRV for `env_filter` in its `Cargo.toml` (see [the comment](https://github.com/rust-cli/env_logger/pull/404#discussion_r3304020726) on crates/env_filter/Cargo.toml) , I thought it may be useful to have a check in the CI to prevent accidental further bumps for the `env_filter` MSRV.

crates/env_filter/src/filter.rs
cagatay-y: At least in my use case, the filter member of the logger is a `OnceCell<env_filter::Filter>` and until the filter is fully initialised, the logger is returned a `Option<&Filter>::None` when trying to `get` it. When it receives the `None`, it simply lets the log message through, so the user is able to see the warning.

crates/env_filter/Cargo.toml
epage: This effectively raises the MSRV of `env_logger` without anything specifically saying so. We need to find an alternative route or bump the MSRV.

cagatay-y: The crate itself does not use the `Error` trait for `ParseError`, so I removed the implementation of the trait for no_std and thus the use of `core::error::Error`. This allows MSRV to stay 1.71.

crates/env_filter/src/parser.rs
cagatay-y: This got resolved with the restoration of the MSRV for the `env_filter` crate.

cagatay-y:
In the latest force-push, in addition to what was raised in the review, I fixed an error I made related to the handling of the features of the `log` dependency in the first version of the PR, and removed the imports for the tests which should have been removed with the change to always have std enabled for the tests.

crates/env_filter/src/filter.rs
epage: > At least in my use case,

While true, I need to consider the general case. Unless there is a reason you feel we should define this to be the general case.

crates/env_filter/Cargo.toml
```diff
@@ -26,12 +26,13 @@ pre-release-replacements = [
 ]
 
 [features]
-default = ["regex"]
+default = ["std", "regex"]
 regex = ["dep:regex"]
+std = ["log/std", "regex/std"]
```
epage: Shouldn't these use `log?/std` / `regex?/std`?

crates/env_filter/src/filter.rs
cagatay-y: My thinking was that it is possible for it to be useful and in the worst case where no logger was set, [the log will be ignored](https://docs.rs/log/latest/log/fn.set_logger.html#:~:text=Any%20log%20events%20that%20occur%20before%20the%20call%20to%20set%5Flogger%20completes%20will%20be%20ignored.), which ends up giving the same result as not logging in the first place. However, I am also completely fine with removing the lines if you think that would be the better approach; or implementing another approach for outputting the error, if you have one in mind.

crates/env_filter/Cargo.toml
cagatay-y: You're right. I have added the question mark for `regex`, but `log?/std` fails because `log` is not an optional dependency.

crates/env_filter/src/filter.rs
epage: I'd prefer doing nothing

cagatay-y: Okay, I've removed it.

epage:
Thanks!

PR #407, opened 2026-05-13
feat(example): show milliseconds since logger init

### LINKED ISSUE #355: Millisecond since init
Log4J has this nice feature that part of the log record is the number of millisecond since the logger initialization. It makes for a very simple timer giving you an idea of how long the process has run. Would it be possible to get it in `env_logger`?

SAY-5:
Adds an example that captures `Instant::now()` and uses it inside the custom format closure to print elapsed milliseconds, matching the approach @vigna confirmed in #355. Closes #355.

### DIFF
--- a/examples/millis_since_init.rs
+++ b/examples/millis_since_init.rs
@@ -1,0 +1,40 @@
+/*!
+Including the milliseconds elapsed since the logger was initialized in each
+record, similar to `Log4J`'s relative time pattern.
+
+Before running this example, try setting the `MY_LOG_LEVEL` environment variable to `info`:
+
+```no_run,shell
+$ export MY_LOG_LEVEL='info'
+```
+*/
+
+use std::io::Write;
+use std::time::{Duration, Instant};
+
+use env_logger::{Builder, Env};
+
+fn init_logger() {
+    let env = Env::default().filter("MY_LOG_LEVEL");
+
+    let start = Instant::now();
+    Builder::from_env(env)
+        .format(move |buf, record| {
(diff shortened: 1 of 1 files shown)


examples/millis_since_init.rs
```diff
@@ -0,0 +1,39 @@
+/*!
```
epage, line 1: *sigh* I was going to ask for this to be switched to line comments but apparently this repo uses multi-line comments already. We'll need to fix that eventually.

examples/millis_since_init.rs
```diff
@@ -0,0 +1,39 @@
...
+            )
+        })
+        .init();
+}
+
+fn main() {
+    init_logger();
+
+    log::info!("a log from `MyLogger`");
+    log::warn!("another log, a moment later");
```
epage, line 39: Should we have sleeps or something else to help make the elapsed time more obvious?

examples/millis_since_init.rs
```diff
@@ -0,0 +1,39 @@
...
+
+use env_logger::{Builder, Env};
+
+fn init_logger() {
+    let env = Env::default().filter("MY_LOG_LEVEL");
+
+    let start = Instant::now();
+    Builder::from_env(env)
+        .format(move |buf, record| {
+            let elapsed = start.elapsed().as_millis();
```
epage, line 23: Should we consolidate this example with `custom_format.rs`?

epage: We can always reevaluate this later.

PR #408, opened 2026-05-21
feat(fmt): support systemd journal priority prefixes

sant0s12:
When `JOURNAL_STREAM` is set, log lines are prefixed with `<N>` (syslog priority) so journald picks up the correct severity.

## API

```rust
use env_logger::{Builder, fmt::JournalPrefix};

let mut builder = Builder::new();
builder.format_journal_prefix(JournalPrefix::Auto);    // default
builder.format_journal_prefix(JournalPrefix::Enable);  // always prefix
builder.format_journal_prefix(JournalPrefix::Disable); // never prefix
```

- **Auto** (default): enables prefixes when `JOURNAL_STREAM` is set, resolved once at logger construction
- **Enable**: always prepend `<N>`
- **Disable**: never prepend

## Priority mapping

| Level | Prefix | Syslog |
|-------|--------|--------|
| Error | `<3>` | LOG_ERR |
| Warn  | `<4>` | LOG_WARNING |
| Info  | `<6>` | LOG_INFO |
| Debug/Trace | `<7>` | LOG_DEBUG |

## Changes

- `src/fmt/mod.rs`: Added `JournalPrefix` enum, `journal_prefix` bool field on `ConfigurableFormat`, resolution in `fmt::Builder::build()`, and `write_journal_prefix()` in the formatter
- `src/logger.rs`: Added `Builder::format_journal_prefix()` method
- Made `use log::Level` unconditional (was behind `#[cfg(feature = "color")]`)

### DIFF
--- a/src/fmt/mod.rs
+++ b/src/fmt/mod.rs
@@ -63,7 +63,6 @@ use std::io::prelude::Write;
 use std::rc::Rc;
 use std::{fmt, io, mem};
 
-#[cfg(feature = "color")]
 use log::Level;
 use log::Record;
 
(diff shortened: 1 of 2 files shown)


epage:
Note that we have an issue at #349 for this where the design is being worked through.

Independent of that, our contribution quality asks for people to work designs out in issues before posting PRs.

As this doesn't match an agreed to design, I'm going to close this for now.

PR #420, opened 2026-09-09
Escape controls without color

### LINKED ISSUE #419: Sanitization requires color, which parses every record
### Problem
With styling disabled, `AutoStream` still walks every record through `anstyle-parse`'s VT state machine, roughly doubling per-record cost even for builds emitting no styling.

`Target::Stderr` → `/dev/null`, 161-byte records, 500k iterations, release + LTO, 8 interleaved rounds:

| | ns/record |
| --- | --- |
| with `color` | 596 |
| without `color` | 302 |

Turning off just `auto-color` doesn't avoid it — `WriteStyle::Auto` then collapses to `Never` (`src/writer/mod.rs:134`), making the parse unconditional. Only dropping `color` entirely removes the parser.

However dropping `color` also removes sanitization, as the two share a code path, so untrusted input then reaches the reader's terminal verbatim:

> Note: deactivating the build-time feature `color` is not a mitigation as that
> removes all ANSI escape code stripping from `env_logger`.

Neither configuration avoids both. The coupling isn't required though: without `color` there is no styling emitted, so nothing needs *parsing*, only payload bytes need neutralizing.

### Proposal
Escape C0 and DEL as `\xNN` when `color` is off, sparing `\n` and `\t`.

- 319 ns/record measured, so sanitizing costs ~17 ns over the unsafe build today
- `color` path untouched, no behavior change for anyone using it
- differs from the strip it replaces in both directions:
  - stricter on `\r`, VT and FF, which `anstream` passes through
  - more permissive on raw C1 and stray continuation bytes, which `anstream` drops (matching it there needs full UTF-8 validation — ~20 lines, free for ASCII records, ~+40 ns/record for records with legitimate non-ASCII)

kskalski:
### What does this PR try to solve?
Closes #419

Escapes C0 and DEL as `\xNN` in `BufferWriter::print` when `color` is off, sparing `\n` and `\t`. Returns `Cow`, so records without control characters are borrowed and never copied.

### Notes to reviewers
Only `#[cfg(not(feature = "color"))]` changes, and that path does nothing today — no behavior change for anyone using `color`.

The fast-path scan folds bitwise rather than using `any`: short-circuiting stops LLVM vectorizing it, costing ~120 ns/record instead of ~17. Commented so it doesn't get simplified back.

Second commit updates the docs warning quoted in #419, which the first makes false.

Verified manually against a payload of CSI, OSC 0/8/52, BEL, CR, DEL and multi-byte UTF-8, under both `--no-default-features` and `--all-features`; existing tests pass in both.

LLM involvement: written with Claude Code (Opus 5) — investigation, implementation, benchmark harness, commit messages — with me directing and reviewing throughout. Figures in #419 were measured from separately built binaries, each sanity-checked for its sanitization behavior before timing.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -218,9 +218,11 @@
 //! - In the application, calling [`Builder::write_style(Never)`][Builder::write_style] to have all ANSI escape codes stripped
 //! - In the application, [stripping ANSI escape codes](https://docs.rs/anstream/latest/anstream/adapter/fn.strip_str.html)
 //!   from user inputs
+//! - Deactivating the build-time feature `color`, which escapes control characters as `\xNN`
+//!   rather than stripping them, and also covers the `\r`, VT and FF that stripping passes on
 //!
-//! Note: deactivating the build-time feature `color` is not a mitigation as that removes all ANSI escape code
-//! stripping from `env_logger`.
+//! Note: `\n` is never escaped, so untrusted input can still read as an extra record. The
+//! default [indentation][Builder::format_indent] offsets such lines by four spaces.
 //!
 //! </div>
 //!
--- a/src/writer/buffer.rs
+++ b/src/writer/buffer.rs
@@ -56,6 +56,8 @@ impl BufferWriter {
         use std::io::Write as _;
 
         let buf = buf.as_bytes();
+        #[cfg(not(feature = "color"))]
+        let buf = &escape_controls(buf);
         match &self.target {
             WritableTarget::WriteStdout => {
                 let stream = io::stdout();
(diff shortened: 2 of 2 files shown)


epage:
This is in violation of our AI policy

kskalski:
Is your policy documented somewhere? Would rewriting it manually or in other way allow the contribution to be accepted? 
I would still like the tradeoff between perf and vulnerability of log files / terminal fixed...

epage:
https://github.com/rust-cli/env_logger/blob/main/AI_POLICY.md

Our contrib guide (and the PR template) say that a decision should be made on Issue before posting PRs

PR #421, opened 2026-09-14
fix: Revert optional feature syntax

KeyWeeUsr:
Hey :wave:

I read [the guide](https://github.com/KeyWeeUsr/env_logger/blob/main/CONTRIBUTING.md) but I'm not sure whether this revert would justify creating a separate issue, so I made a PR instead. Feel free to close/convert if needed.

### What does this PR try to solve?
I'm experiencing an issue when vendoring caused by https://github.com/rust-cli/env_logger/commit/132fe86c8cb8e5df4fca7d71067a8d862a366b95 ([`package-name?/feature`](https://doc.rust-lang.org/cargo/reference/features.html?highlight=package-name?#dependency-features) specifically). I'm not sure whether this is a bug in Cargo potentially opened elsewhere, however I reverted the problematic parts locally this way. The commit also contains a POC which might also be useful as a CI test for vendoring in case such change happens again in the future, so it's caught before releasing.

POC below:

    FROM rust:alpine
    WORKDIR /tmp

    RUN apk add git
    RUN git clone --depth=1 https://github.com/rust-lang/log vendor/log
    RUN git clone --depth=1 https://github.com/rust-cli/env_logger vendor/env_logger

    RUN <<EOR
    cat<<EOF > vendor/env_logger/fix.patch
    diff --git a/crates/env_filter/Cargo.toml b/crates/env_filter/Cargo.toml
    --- a/crates/env_filter/Cargo.toml
    +++ b/crates/env_filter/Cargo.toml
    @@ -28,11 +28,11 @@ pre-release-replacements = [
     [features]
     default = ["std", "regex"]
     regex = ["dep:regex"]
    -std = ["log/std", "regex?/std"]
    +std = ["log/std"]

     [dependencies]
     log = { version = "0.4.29" }
    -regex = { version = "1.12.3", optional = true, default-features=false, features=["perf"] }
    +regex = { version = "1.12.3", optional = true, default-features=false, features=["std", "perf"] }

     [dev-dependencies]
     snapbox = "1.0"
    EOF

    cd vendor/env_logger&&git apply fix.patch
    EOR

    RUN cat vendor/env_logger/crates/env_filter/Cargo.toml
    RUN <<EOR
    mkdir src
    echo "fn main() {}" > src/main.rs

    cat<<EOF > Cargo.toml
    [package]
    name = "bug"
    edition = "2024"

    [dependencies]
    env_logger = { version = "0.11.11", default-features = false, features = [] }
    log = { version = "0.4.34", default-features = false }
    # note: happens even if not specified here
    env_filter = { version = "2.0.0", default-features = false, features = [] }

    [patch.crates-io]
    env_logger = { path = "vendor/env_logger" }
    env_filter = { path = "vendor/env_logger/crates/env_filter" }
    log = { path = "vendor/log" }
    EOF
    EOR

    RUN <<EOR
    rustc --version
    cargo --version
    cargo build --offline --no-default-features --verbose
    EOR

### Notes to reviewers

POC above and in the commit message.
LLM involvement: Zero


epage:
I can't imagine what this would be the root cause for, rather than a workaround. I'd prefer to focus on the root cause. The only vendor issue I'm aware of is https://github.com/rust-lang/cargo/issues/10801.

Going to close this in favor of whereever the root cause Issue is opened.

Note that If reverting somethìng, then likely this isn'	t so straightforward that you can skip an issue. An Issue would have better encouraged describing the problem, more than the PR template does.

KeyWeeUsr:
@epage I believe that's the bug I'm experiencing! Thanks

